PR #19, opened 2018-12-14
Set io_queue for some requests

Donaim:
Propagate id of ocf_queue where it is possible

This resolves the issue of
  bottom adapter always getting an ocf_io with io_queue = 0,
  no matter from which queue function to bottom adapter was called

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/19)


Donaim:
I've tested this on SPDK CAS-bdev implementation - every queue_id is propagated correctly.

PR #53, opened 2019-02-01
Update metadata_test_dirty_all to metadata_test_dirty_sec to avoid un…

MinneyJohn:
When determine whether we need to flush metadata for one write IO in write back mode,

here should be the design point:

If the new write IO will NOT introduce new dirty sector, there is NO need to flush the metadata; Because the existing metadata already have enough information to track the dirty data
If the new write IO will introduce new dirty sectors, then we need to flush the metadata to track the new dirty sectors on the cache drives.
Our current OCF code tries to match the design point, but there is a small hole.

We rely on "dirty_all" to check whether all the sectors of the request are dirty.

But now we check the dirty bits for all sectors of each mapped cache line.

For the first and last core line in the request, this may be wrong.

Because, the first and last core line in the request may be NOT complete. For example, the request may want to get only 8th sector of the first core line (block) and the 1st sector of the last core line (block).

So for the first and last core line, we should only check the dirty bits of corresponding sectors requested by the write IO.

Or we'll introduce unnecessary metadata flush and reduce the write performance.

So use metadata_test_dirty_sec to check each cache line.
And for the first and last core line, maybe partial cache line will be checked.
For other core lines, the whole cache line will be checked.

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/53)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ utility.
 * [Source Code](#source)
 * [Deployment](#deployment)
 * [Unit Tests](#tests)
-* [Unit Tests](#Demo Build)
 * [Contributing](#contributing)
 
 <a id="documentation"></a>
@@ -67,18 +66,7 @@ To run unit test you need to install following packages:
 To run unit tests use following command:
 
 ~~~{.sh}
-./tests/unit/framework/run_unit_tests.py
-~~~
-
-<a id="Demo Build"></a>
-## Demo Build
-
-OCF is shipped with some simple demo to compile and use OCF.  
-You can try to build this demo to verify that you change can pass the compile.
-
-~~~{.sh}
-cd ./example/simple/
-make
+./tests/ut-framework/run_unit_tests.py
 ~~~
 
 <a id="contributing"></a>
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ utility.
 * [Source Code](#source)
 * [Deployment](#deployment)
 * [Unit Tests](#tests)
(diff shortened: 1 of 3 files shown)


MinneyJohn:
> Actually `dirty_all` is used to determine if we need to check if flush is needed. We didn't do unnecessary flush, but we did unnecessary check and your patch fixes this problem. I'm going to accept it, but please update the commit message.
> 
> > _[Reviewable](https://reviewable.io/reviews/open-cas/ocf/53#-:-LXcVekm3x7nK_nocLUB:b-w8ijbq)_ status: 0 of 2 LGTMs obtained
> 
> _[src/engine/engine_common.c, line 120 at r1](https://reviewable.io/reviews/open-cas/ocf/53#-LXcZ0QdDbd7NKyoALvq:-LXcZ0QeCLVeVAh4U0NV:b-jy934e) ([raw file](https://github.com/open-cas/ocf/blob/4e3fa6878baef26ace5983048b6db35d79232b0b/src/engine/engine_common.c#L120)):_
> 
> > ```c
> > 			/* Check if cache line is fully dirty */
> > 			if (metadata_test_dirty_sec(cache, _entry->coll_idx, start_sector, end_sector))
> > ```
> 
> Please break this line to not exceed 80 characters.

Did more discussion with Robert, this code issue do introduce unnecessary metadata update.
And also find some other places which may need improvement.
So I'll close this pull request and submit new pull request with more update.

PR #69, opened 2019-03-08
Alru clean improvement

mmichal10:
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/69)

### DIFF
--- a/src/cleaning/alru.c
+++ b/src/cleaning/alru.c
@@ -51,8 +51,7 @@ struct flush_merge_struct {
 
 struct alru_flush_ctx {
 	struct ocf_cleaner_attribs attribs;
-	struct ocf_user_part *parts[OCF_IO_CLASS_MAX];
-	int part_id;
+	bool flush_perfomed;
 	uint32_t clines_no;
 	ocf_cache_t cache;
 	ocf_cleaner_end_t cmpl;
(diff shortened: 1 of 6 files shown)


robertbaldyga:
___
*[src/cleaning/alru.c, line 812 at r2](https://reviewable.io/reviews/open-cas/ocf/69#-L_S6-L5CGUbz25vqEps:-L_S6-L5CGUbz25vqEpt:b9x67xg) ([raw file](https://github.com/open-cas/ocf/blob/c8669866e7a761f5407dedd6df8ad00641272827/src/cleaning/alru.c#L812)):*
> ```C
> 
> 	fctx->attribs.cmpl_context = fctx;
> 	fctx->attribs.cmpl_fn = NULL;
> ```

Huh? Who calls cleaner completion function after cleaning is completed? We don't want to complete immediately if cleaning is performed - if it takes long time, it's possible that second cleaner iteration will be started before the previous one has been finished. This can happen multiple times, which would make cleaner blocked for good (stopping the cache would be impossible until all dirty data will be cleaned, which may take a looong time).

robertbaldyga:
___
*[src/cleaning/alru.c, line 793 at r2](https://reviewable.io/reviews/open-cas/ocf/69#-L_S7UUt1vX_DFsPK8jM:-L_S7UUt1vX_DFsPK8jN:b-pg2psy) ([raw file](https://github.com/open-cas/ocf/blob/c8669866e7a761f5407dedd6df8ad00641272827/src/cleaning/alru.c#L793)):*
> ```C
> 	flush_performed = alru_do_clean(cache, fctx);
> 
> 	OCF_REALLOC_DEINIT(&fctx->flush_data, &fctx->flush_data_limit);
> ```

This deinit may be called before flushing iteration is finished. Please move it to completion function.

robertbaldyga:
___
*[src/cleaning/alru.c, line 783 at r2](https://reviewable.io/reviews/open-cas/ocf/69#-L_S85sACpyc2wQNbill:-L_S85sACpyc2wQNbilm:bazum61) ([raw file](https://github.com/open-cas/ocf/blob/c8669866e7a761f5407dedd6df8ad00641272827/src/cleaning/alru.c#L783)):*
> ```C
> }
> 
> static void alru_clean(void *priv, int error)
> ```

We could merge this function with alru_do_clean, as after removing per-partition flushing it became very simple.

robertbaldyga:
___
*[src/cleaning/nop.c, line 12 at r2](https://reviewable.io/reviews/open-cas/ocf/69#-L_SALVy-1MafEj_HZGk:-L_SALVy-1MafEj_HZGl:b-quq2as) ([raw file](https://github.com/open-cas/ocf/blob/c8669866e7a761f5407dedd6df8ad00641272827/src/cleaning/nop.c#L12)):*
> ```C
> void cleaning_nop_perform_cleaning(ocf_cache_t cache, ocf_cleaner_end_t cmpl)
> {
> 	uint32_t interval = 20000;
> ```

Why don't we make this value huge? We don't need to wake up cleaner thread every 20s for nop policy. The only reason for this value I can imagine is that we want to apply change of cleaning policy in reasonable time. Maybe let's fix design instead and introduce kick() method for cleaner thread. It would be called on cleaning policy on params change, so that all updates will have immediate effect. What do you think? It doesn't have to be done it this patchset, but it would be a nice thing to have.

robertbaldyga:
___
*[src/cleaning/alru.c, line 799 at r5](https://reviewable.io/reviews/open-cas/ocf/69#-L_T4UMn4GgUfX9dLm7P:-L_T4UMn4GgUfX9dLm7Q:bs8ruwp) ([raw file](https://github.com/open-cas/ocf/blob/fced3c02c24eabb78abbd6117d922cc6a4e1a65b/src/cleaning/alru.c#L799)):*
> ```C
> 		OCF_METADATA_UNLOCK_WR();
> 		return;
> 	} else {
> ```

Unnecessary 'else' after 'return'. (No-else-return)

PR #71, opened 2019-03-11
Make management API asynchronous

robertbaldyga:
NOTE: This patch only changes API that pretends to be asynchronous.
Most of management operations are still performed synchronously.
The real asynchronism will be introduced in the next patches.

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/71)

### DIFF
--- a/inc/ocf_core.h
+++ b/inc/ocf_core.h
@@ -56,17 +56,6 @@ static inline const struct ocf_volume_uuid *ocf_core_get_uuid(ocf_core_t core)
 }
 
 /**
- * @brief Asociate new UUID value with given core
- *
- * @param[in] core Core object
- * @param[in] uuid new core uuid
- *
- * @retval 0 Success
- * @retval Non-zero Fail
- */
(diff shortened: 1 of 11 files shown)


robertbaldyga:
___
*[src/mngt/ocf_mngt_cache.c, line 1653 at r1](https://reviewable.io/reviews/open-cas/ocf/71#-L_hQ10z5IKkfkxSVvm8:-L_iIsAFE7p74J51w7wc:b49l5pa) ([raw file](https://github.com/open-cas/ocf/blob/e21b617841f39f600f9ae41b2353ec274d9e0649/src/mngt/ocf_mngt_cache.c#L1653)):*
<details><summary><i>Previously, mmichal10 (Michał Mielewczyk) wrote…</i></summary><blockquote>

Why did you remove _ocf_mngt_init_handle_error() call?
</blockquote></details>

To make API symmetric. It seems natural that cache is initlialized by `ocf_mngt_cache_start()` and deinitialized by `ocf_cache_stop()`. Performing deinitialization in `ocf_mngt_cache_load()` error path breaks this symmetry, and leads to behavior that is probably not expected by user.

PR #73, opened 2019-03-14
Add kick function for cleaner

mmichal10:
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/73)

### DIFF
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -1347,6 +1347,15 @@ static void _ocf_mngt_attach_init_instance(ocf_pipeline_t pipeline,
 {
 	struct ocf_cache_attach_context *context = priv;
 	ocf_cache_t cache = context->cache;
+	int result;
+
+	result = ocf_start_cleaner(cache);
+	if (result) {
+		ocf_cache_log(cache, log_err,
+				"Error while starting cleaner\n");
+		OCF_PL_FINISH_RET(context->pipeline, result);
+	}
+	context->flags.cleaner_started = true;
 
 	switch (cache->device->init_mode) {
 	case ocf_init_mode_init:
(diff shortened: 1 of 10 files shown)


michalwy:
retest this please

Deixx:
retest this please

src/mngt/ocf_mngt_flush.c
```diff
@@ -974,6 +974,9 @@ int ocf_mngt_cache_cleaning_set_policy(ocf_cache_t cache, ocf_cleaning_t type)
 
 	ocf_metadata_unlock(cache, OCF_METADATA_WR);
 
+	if (type != ocf_cleaning_nop)
+		ocf_kick_cleaner(cache);
```
michalwy: I would move this to cleaning policy initialize() op. It's cleaning policy responsibility to decide whether it wants cleaner to be kicked or not.

mmichal10: Moving it there would cause a problem because in cache attach procedure initialize() is called before cleaner thread is started.

robertbaldyga: @mmichal10 
What about starting cleaner earlier? We can define the rule, that ocf_cleaner_run() shall not be called before first kick. Besides, even if it was, the thread would immediately go to sleep because the cache is locked, so nothing wrong could happen.

src/mngt/ocf_mngt_cache.c
```diff
@@ -1347,6 +1347,18 @@ static void _ocf_mngt_attach_init_instance(ocf_pipeline_t pipeline,
...
 	struct ocf_cache_attach_context *context = priv;
 	ocf_cache_t cache = context->cache;
+	int result;
+
+	if (!context->flags.cleaner_started) {
+		result = ocf_start_cleaner(cache);
+		if (result) {
+			ocf_cache_log(cache, log_err,
+					"Error while starting cleaner\n");
+			ocf_pipeline_finish(context->pipeline, result);
```
robertbaldyga:
```
OCF_PL_FINISH_RET(pipeline, result);
```

src/mngt/ocf_mngt_cache.c
```diff
@@ -1347,6 +1347,18 @@ static void _ocf_mngt_attach_init_instance(ocf_pipeline_t pipeline,
 {
 	struct ocf_cache_attach_context *context = priv;
 	ocf_cache_t cache = context->cache;
+	int result;
+
+	if (!context->flags.cleaner_started) {
```
robertbaldyga: Is it possible that cleaner is somehow started earlier? I see this is the only function where it's started, which means that it would have to be called twice during attach procedure, which is not the case. Hence I think this check is unnecessary.

src/mngt/ocf_mngt_cache.c
```diff
@@ -1347,6 +1347,18 @@ static void _ocf_mngt_attach_init_instance(ocf_pipeline_t pipeline,
...
+		result = ocf_start_cleaner(cache);
+		if (result) {
+			ocf_cache_log(cache, log_err,
+					"Error while starting cleaner\n");
+			ocf_pipeline_finish(context->pipeline, result);
+			return;
```
micrakow: please use OCF_PL_FINISH_RET macro

PR #75, opened 2019-03-15
Implement asynchronous attach, load and stop

robertbaldyga:
NOTE: This is still not the real asynchronism. Metadata interfaces
are still not fully asynchronous.

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/75)

### DIFF
--- a/inc/ocf_cache.h
+++ b/inc/ocf_cache.h
@@ -266,12 +266,4 @@ void ocf_cache_set_priv(ocf_cache_t cache, void *priv);
  */
 void *ocf_cache_get_priv(ocf_cache_t cache);
 
-/**
- * @brief Set queue to be used during flush operation
- *
- * @param[in] cache Cache object
- * @param[in] queue Queue object
- */
-void ocf_cache_set_flush_queue(ocf_cache_t cache, ocf_queue_t queue);
-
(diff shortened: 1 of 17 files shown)


robertbaldyga:
___
*[src/mngt/ocf_mngt_cache.c, line 1509 at r2](https://reviewable.io/reviews/open-cas/ocf/75#-LaFvWm6BIdT034bwK2_:-LaGAzB916wAGNEL4stx:birk7kq) ([raw file](https://github.com/open-cas/ocf/blob/7e7c9ca8d62805796046beb52757cdd343ce9b45/src/mngt/ocf_mngt_cache.c#L1509)):*
<details><summary><i>Previously, arutk (Adam Rutkowski) wrote…</i></summary><blockquote>

You can embed pipeline struct as a member of context and use "container_of" to get context pointer 
</blockquote></details>

I planned to embed context as a `struct pipeline` member (by taking sizeof() of context for pipeline construction). That way pipeline would take care of context allocation. In that scenario all callbacks would be able to take pipeline as `priv`, so indeed with that (and a bit of callback function definition unification) the generic handling of simple completion will be possible. Let's just mark it as TBD in future.

PR #112, opened 2019-04-10
Tests for flush method

katlapinka:
Added tests for flush method and placeholders for missing pyocf methods

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/112)


KamilLepek:
retest this please

tests/functional/tests/conftest.py
```diff
@@ -3,18 +3,19 @@
...
 import pytest
 
 sys.path.append(os.path.join(os.path.dirname(__file__), os.path.pardir))
+
 from pyocf.types.logger import LogLevel, DefaultLogger, BufferLogger
 from pyocf.types.volume import Volume, ErrorDevice
 from pyocf.types.ctx import get_default_ctx
-
-
-def pytest_configure(config):
```
imjfckm: Doesn't seem like a good idea to remove that, if someone hasn't got pyocf on env PATH/PYTHONPATH the tests won't find it.

tests/functional/tests/conftest.py
```diff
@@ -37,3 +38,18 @@ def pyocf_ctx_log_buffer():
     yield logger
     for cache in c.caches:
         cache.stop()
+
+
+@pytest.fixture()
+def start_cache_with_core():
+    def start_with_config(cache_size: Size, core_size: Size, **config):
```
imjfckm: Is this a fixture that returns a function to call when wanting to start cache/core pair? Consider putting it in some test utils (I believe there is already a tests.utils module), I'm not sure writing this as a fixture is a good pattern.

tests/functional/tests/flush/flush_test.py
```diff
@@ -0,0 +1,289 @@
...
+    assert cmpl.results["err"] == 0
+
+
+def check_configuration(cache, expected_conf):
+    stats = cache.get_stats()
+    assert stats["conf"]["cache_line_size"] == expected_conf["cache_line_size"]
+    assert stats["conf"]["cache_mode"] == expected_conf["cache_mode"]
+
+
+def check_statistics(stats, expected_occupancy, expected_dirty, expected_clean):
```
imjfckm: This function's name doesn't match its purpose. It's named "check statistics" when in fact it only checks if occupancy/dirty/clean values are as expected. What's wrong with having all those assertions written explicitly in tests which need them?

tests/functional/tests/flush/flush_test.py
```diff
@@ -0,0 +1,289 @@
...
+# TODO: remove xfail when wb cache mode will be fixed
+@pytest.mark.xfail
+@pytest.mark.parametrize("cls", CacheLineSize)
+def test_wb_flush_cache(pyocf_ctx, start_cache_with_core, cls):
+    cache_device, core_device, cache, core = start_cache_with_core(Size.from_MiB(30),
+                                                                   Size.from_MiB(60),
+                                                                   cache_mode=CacheMode.WB,
+                                                                   cache_line_size=cls)
+
+    check_configuration(cache, {"cache_line_size": cls, "cache_mode": CacheMode.WB})
```
imjfckm: Do we need to check if the cache we just started has the configuration we wanted? I think it should be already tested in start/stop tests and here it's only noise.
Same for all the other tests

tests/functional/tests/flush/flush_test.py
```diff
@@ -0,0 +1,289 @@
...
+    stats_after_flush = cache.get_stats()
+    check_statistics(stats_after_flush,
+                     stats_before_flush["usage"]["occupancy"]["value"],
+                     0,
+                     stats_before_flush["usage"]["clean"]["value"] + math.ceil(io_size.B / 4096))
+
+    assert core.exp_obj_md5() == core_device.md5()
+
+    cache.remove_core(core)
+    cache.stop()
```
imjfckm: Context cleanup takes care of leftover cache and cores. If you don't test it, you probably don't need to call stop/remove_core/etc

tests/functional/tests/conftest.py
katlapinka: done

tests/functional/tests/flush/flush_test.py
katlapinka: Done

tests/functional/tests/flush/flush_test.py
katlapinka: Done

tests/functional/tests/flush/flush_test.py
katlapinka: Done

arutk:
Closing pull request until reworked by the author.

PR #125, opened 2019-04-17
Wait for IO put in volume close

arutk:
Volume close should not close underlying device until all
I/O targeting this volume are deallocated. To achieve this
a reference counter is added to volume. Counter value
matches number of I/O objects associated with volume. Counter
is freezed when volume is closed, blocking allocation of new
I/O objects.

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/125)

### DIFF
--- a/src/ocf_io.c
+++ b/src/ocf_io.c
@@ -54,9 +54,14 @@ struct ocf_io *ocf_io_new(ocf_volume_t volume)
 	struct ocf_io_meta *io_meta;
 	void *data;
 
-	data = env_allocator_new(volume->type->allocator);
-	if (!data)
+	if (!ocf_refcnt_inc(&volume->refcnt))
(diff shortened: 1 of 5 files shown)


src/ocf_io.c
```diff
@@ -54,6 +54,9 @@ struct ocf_io *ocf_io_new(ocf_volume_t volume)
 	struct ocf_io_meta *io_meta;
 	void *data;
 
+	if (!ocf_refcnt_inc(&volume->refcnt))
+		return NULL;
+
 	data = env_allocator_new(volume->type->allocator);
 	if (!data)
 		return NULL;
```
michalwy: We need to decrease refcnt in error path

arutk: done

PR #126, opened 2019-04-17
Rename 'printf' field name in logger ops to 'print'.

mmichal10:
Due the aggresive security checks in compiler 'printf' might be substituded with
'__printf_chk'. However it does not differentiate whether substituted string is
library function call whether filed in structure.

By renaming filed we prevent it to be unintentionally subustituted by the
preprocessor.

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/126)

### DIFF
--- a/inc/ocf_logger.h
+++ b/inc/ocf_logger.h
@@ -31,9 +31,9 @@ typedef enum {
 struct ocf_logger_ops {
 	int (*open)(ocf_logger_t logger);
 	void (*close)(ocf_logger_t logger);
-	int (*printf)(ocf_logger_t logger, ocf_logger_lvl_t lvl,
+	int (*print)(ocf_logger_t logger, ocf_logger_lvl_t lvl,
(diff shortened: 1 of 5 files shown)


inc/ocf_logger.h
```diff
@@ -31,7 +31,7 @@ typedef enum {
 struct ocf_logger_ops {
 	int (*open)(ocf_logger_t logger);
 	void (*close)(ocf_logger_t logger);
-	int (*printf)(ocf_logger_t logger, ocf_logger_lvl_t lvl,
+	int (*print)(ocf_logger_t logger, ocf_logger_lvl_t lvl,
 			const char *fmt, va_list args);
 	int (*printf_rl)(ocf_logger_t logger, const char *func_name);
```
michalwy: I would change also printf_rl() to print_rl() to have consistent naming(). Right now we have print() and print**f**_rl()

PR #133, opened 2019-04-19
Ajrutkow async counters

arutk:


inc/ocf_err.h
```diff
@@ -95,6 +95,25 @@ typedef enum {
 
 	/** Invalid cache line size */
 	OCF_ERR_INVALID_CACHE_LINE_SIZE,
+
+	/** Unable to perform operation due to concurrent management request */
+	OCF_ERR_MNGMT_OP_IN_PROGRESS,
```
michalwy: We need to have consistent naming for management related things. Sometime we use mngmt, sometimes mngt. We need to have one prefix.

arutk: fixed this one, will add backlog item to track fixing of other instances.

src/utils/utils_req.c
```diff
@@ -319,7 +338,8 @@ void ocf_req_clear_map(struct ocf_request *req)
 			   sizeof(req->map[0]) * req->core_line_count, 0));
 }
 
+/* TODO: get rid of this and use async wait on counter  */
 uint32_t ocf_req_get_allocated(struct ocf_cache *cache)
 {
-	return env_atomic_read(&cache->pending_requests);
+	return env_atomic_read(&cache->refcnt.io_req.counter);
```
michalwy: I think it would be better to have ocf_refcnt_get() or similar getter instead of accessing refcnt fields directly. In general all operations on refcnt should be done using refcnt API, not using direct access.

arutk: This code is removed in one of the following patches. The function to read current value is not needed anywhere in the code.

src/utils/utils_req.h
```diff
@@ -92,15 +92,6 @@ int ocf_req_new_extended(struct ocf_request **out_req, ocf_queue_t queue,
...
 		ocf_core_t core, uint64_t addr, uint32_t bytes, int rw);
 
-/**
- * @brief Get number of allocated requests
- *
- * @param cache OCF cache instance
- *
- * @return Number of allocated requests
- */
-uint32_t ocf_req_get_allocated(struct ocf_cache *cache);
```
robertbaldyga: I'm not sure if SPDK doesn't need it for their OCF integration. We need to confirm it before we remove this function from our API.

arutk: SPDK team confirmed they won't be needing this. At the moment of stopping cache top bdevs are closed and have all IOs completed. Bottom bdevs are still open and can receive I/O. Outstanding backfill requests are drained in detach step where we freeze and wait for metadata ref counter.

tests/unit/tests/utils/utils_refcnt.c/utils_refcnt_register_zero_cb.c
```diff
@@ -0,0 +1,87 @@
...
+#include <stddef.h>
+#include <setjmp.h>
+#include <cmocka.h>
+#include "print_desc.h"
+
+#include "../utils/utils_refcnt.h"
+
+static void zero_cb(void *ctx)
+{
+	*(volatile int *)ctx = 1;
```
robertbaldyga: Use function_called() instead. We should also check if passed argument is correct by doing check_expected().

arutk: done

PR #134, opened 2019-04-19
Fix immediate reassign

Donaim:
Don't reassign value of cache without any previus use.
It produced warnings when analyzing with scanbuild.


michalwy:
ok to test

PR #140, opened 2019-05-02
Fix example

### LINKED ISSUE #139: Sample compile error

robertbaldyga:
Resolves #139

### DIFF
--- a/example/simple/src/main.c
+++ b/example/simple/src/main.c
@@ -141,12 +141,14 @@ int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
 	 * or adding core object.
 	 */
 	ret = ocf_queue_create(*cache, &cache_priv->mngt_queue, &queue_ops);
-	if (ret)
-		goto err_cache;
+	if (ret) {
+		ocf_mngt_cache_stop(*cache, simple_complete, &context);
+		goto err_priv;
+	}
 
 	/*
 	 * Assign management queue to cache. This has to be done before any
-	 * other management operation. Management queue it treated specially,
+	 * other management operation. Management queue is treated specially,
 	 * and it may not be used for submitting IO requests. It also will not
 	 * be put on the cache stop - we have to put it manually at the end.
 	 */
@@ -154,9 +156,8 @@ int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
 
 	/* Create queue which will be used for IO submission. */
 	ret = ocf_queue_create(*cache, &cache_priv->io_queue, &queue_ops);
-	if (ret) {
-		goto err_queue;
-	}
+	if (ret)
+		goto err_cache;
 
 	/* Attach volume to cache */
 	ocf_mngt_cache_attach(*cache, &device_cfg, simple_complete, &context);
@@ -167,7 +168,6 @@ int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
 
 err_cache:
 	ocf_mngt_cache_stop(*cache, simple_complete, &context);
-err_queue:
 	ocf_queue_put(cache_priv->mngt_queue);
 err_priv:
 	free(cache_priv);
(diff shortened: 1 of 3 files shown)


fanlonglong8500:
i make the example work myself and i found same problem with this, could you  help me with your answer?
1.The pipeline implemented by the queue will result in a very long call stack
2. Why cache mngt_queue need to be created by user?I think this queue is best created in the cache.

robertbaldyga:
Hi @fanlonglong8500 
1. Queues are designed to be handled asynchronously, so normally ocf_queue_run() would be called in separate thread or at least in event loop. That would prevent creating long call stack.
In this example ocf_queue_run() is called synchronously for simplicity.
2. We consider changing it in the future (queue ops cloud be passed for example via cache_cfg), but we also consider sharing queues between caches. In the second option the queue has to be created by user. Initially we decided to make queue creation process consistent for all the queues, which leaves us the way open to choose any of those two options.
In any case, the queue will still have to be destroyed by the user after stopping the cache, because cache stop procedure is performed in mngt_queue execution context, and the queue cannot be destroyed from its own execution context.

example/simple/src/main.c
```diff
@@ -245,6 +356,13 @@ int main(int argc, char *argv[])
 	if (initialize_cache(ctx, &cache1))
 		error("Unable to start cache\n");
 
+	/*
+	 * Increase cache refcount. Otherway cache will be destroyed
+	 * on cache stop preventing destroying of management queue,
+	 * and thus leading to memory leak.
+	 */
+	ocf_mngt_cache_get(cache1);
```
arutk: Let's integrate https://github.com/Open-CAS/ocf/pull/144 first and remove this get/put

robertbaldyga: Right! :)

fanlonglong8500:
Hi @robertbaldyga 
Thank you for explaining these to me.

robertbaldyga:
@fanlonglong8500 
You're welcome!

example/simple/src/main.c
```diff
@@ -81,27 +118,79 @@ int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
...
+	 */
+	ret = ocf_queue_create(*cache, &cache_priv->mngt_queue, &queue_ops);
+	if (ret)
+		goto err_cache;
 
-	/* Attach volume to cache */
-	ret = ocf_mngt_cache_attach(*cache, &device_cfg);
+	/*
+	 * Assign management queue to cache. This has to be done before any
+	 * other management operation. Management queue it treated specially,
```
micrakow: Management queue *is* treated

example/simple/src/main.c
```diff
@@ -81,27 +118,79 @@ int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
...
+	 * be put on the cache stop - we have to put it manually at the end.
+	 */
+	ocf_mngt_cache_set_mngt_queue(*cache, cache_priv->mngt_queue);
+
+	/* Create queue which will be used for IO submission. */
+	ret = ocf_queue_create(*cache, &cache_priv->io_queue, &queue_ops);
 	if (ret) {
-		ocf_mngt_cache_stop(*cache);
-		return ret;
+		goto err_queue;
```
micrakow: err_cache

example/simple/src/main.c
```diff
@@ -81,27 +118,79 @@ int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
...
 	}
 
+	/* Attach volume to cache */
+	ocf_mngt_cache_attach(*cache, &device_cfg, simple_complete, &context);
+	if (ret)
+		goto err_cache;
```
micrakow: what about putting io_queue in case of err here?

robertbaldyga: The io queue will be put during cache stop anyway, so it would be reduntant.

example/simple/src/main.c
robertbaldyga: Done.

example/simple/src/main.c
robertbaldyga: Done.

PR #143, opened 2019-05-06
Fix loading of cache in pyocf

imjfckm:
Flush/load metadata paths are heavily dependend on Data behaving
correctly in terms of seeks/position and that needed to be fixed.

### DIFF
--- a/tests/functional/pyocf/types/cache.py
+++ b/tests/functional/pyocf/types/cache.py
@@ -191,11 +191,10 @@ class Cache:
             self.cache_handle, cache_mode
         )
 
-        if status:
-            self.put_and_write_unlock()
-            raise OcfError("Error changing cache mode", status)
-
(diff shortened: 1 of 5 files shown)


tests/functional/pyocf/types/cache.py
```diff
@@ -472,6 +477,23 @@ def get_default_queue(self):
...
+
+        self.get_and_write_lock()
+        c = OcfCompletion(
+            [("cache", c_void_p), ("priv", c_void_p), ("error", c_int)]
+        )
+        self.owner.lib.ocf_mngt_cache_save(self.cache_handle, c, None)
+
+        c.wait()
+        if c.results["error"]:
+            self.put_and_write_unlock()
```
arutk: you can put_and_write_unlock() unconditionally and then check for error

PR #146, opened 2019-05-08
Initialize and deinitialize core pool with ctx

robertbaldyga:

### DIFF
--- a/src/mngt/ocf_mngt_core_pool_priv.h
+++ b/src/mngt/ocf_mngt_core_pool_priv.h
@@ -1,0 +1,15 @@
+/*
+ * Copyright(c) 2012-2018 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause-Clear
+ */
+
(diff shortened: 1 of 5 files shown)


src/mngt/ocf_mngt_core_pool_priv.h
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright(c) 2012-2018 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause-Clear
+ */
+
+#ifndef __OCF_MNGT_CORE_POOL_PRIV_H__
+#define __OCF_MNGT_CORE_POOL_PRIV_H__
+
+#include "ocf/ocf.h"
+
+void ocf_mngt_core_pool_init(ocf_ctx_t ctx);
```
micrakow: Do we really need this since core_pool_init/deinit are already declared in ocf_mngt.h?

robertbaldyga: Fixed. Now it's not. As OCF does it automatically, it no longer needs to be part of public API.

PR #149, opened 2019-05-09
Set core_id during load if user did not specify one

Donaim:
If cfg->core_id is OCF_CORE_MAX and that core matches the UUID of
existing not opened one, then set cfg->core_id to id of that core.

This is useful when loading cache from metadata: if user does not store
the ids of cores but relies on OCF to assign them, there is no need to
not reassign them on load.

Previus behaviur when cfg->core_id != id of core with matching UUID is
maintained.


Donaim:
This change is useful for SPDK implementation as we don't want to serialize the IDs ourselves

michalwy:
ok to test

PR #152, opened 2019-05-10
Fix setting of IO Data offset

imjfckm:

### DIFF
--- a/tests/functional/pyocf/types/io.py
+++ b/tests/functional/pyocf/types/io.py
@@ -108,9 +108,8 @@ class Io(Structure):
             byref(self), addr, length, direction, io_class, flags
         )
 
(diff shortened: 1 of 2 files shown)


tests/functional/pyocf/types/io.py
```diff
@@ -108,9 +108,8 @@ def configure(
             byref(self), addr, length, direction, io_class, flags
         )
 
-    def set_data(self, data: Data):
-        self.data = data
-        OcfLib.getInstance().ocf_io_set_data_wrapper(byref(self), data, 0)
+    def set_data(self, data: Data, offset=0):
```
Deixx: submitting io after setting non-zero offset ends in Exception

PR #153, opened 2019-05-13
metadata: Put io before callback invocation

robertbaldyga:
This allows for safe volume destruction in callback function.


Donaim:
This patch is very useful for SPDK implementation! Without it, we would need to free volume in concurent manner which is very confusing for such a straightforward thing.

PR #159, opened 2019-05-20
Store core handle instead of core_id in ocf_request

robertbaldyga:


src/engine/engine_bf.c
```diff
@@ -51,25 +51,25 @@ static void _ocf_backfill_complete(struct ocf_request *req, int error)
...
+	ctx_data_free(cache->owner, req->data);
+	req->data = NULL;
+
+	if (req->error) {
+		env_atomic_inc(&req->core->counters->cache_errors.write);
+		ocf_engine_invalidate(req);
+	} else {
+		ocf_req_unlock(req);
+
+		/* always free the request at the last point
```
arutk: Can you move more text to this line to fill 80 char width?

src/engine/engine_common.c
```diff
@@ -24,7 +24,7 @@ void ocf_engine_error(struct ocf_request *req,
 	if (stop_cache)
 		env_bit_clear(ocf_cache_state_running, &cache->cache_state);
 
-	ocf_core_log(&cache->core[req->core_id], log_err,
+	ocf_core_log(cache->core, log_err,
```
arutk: The first argument should be "req->core"

robertbaldyga: Done!

src/engine/engine_bf.c
robertbaldyga: Done. After some rephrasing it actually fits into 80 chars.

PR #168, opened 2019-05-23
Introduce helpers for setting config structures to defaults

robertbaldyga:


src/mngt/ocf_mngt_cache.c
```diff
@@ -1594,7 +1594,7 @@ static int _ocf_mngt_cache_validate_device_cfg(
 	if (device_cfg->uuid.size > OCF_VOLUME_UUID_MAX_SIZE)
 		return -OCF_ERR_INVAL;
 
-	if (device_cfg->cache_line_size &&
+	if (device_cfg->cache_line_size != ocf_cache_line_size_none &&
```
imjfckm: Why separate the check for none and `ocf_cache_line_size_is_valid`? The `is_valid` will return return false for `ocf_cache_line_size_none` and that should be enough

arutk: ocf_cache_line_size_none is valid only in single API function (attach) - in multiple other places "none" is considered invalid.

PR #171, opened 2019-05-24
Associate core metadata with core object

robertbaldyga:


src/utils/utils_core.h
```diff
@@ -6,8 +6,5 @@
 #ifndef __UTILS_CORE_H__
 #define __UTILS_CORE_H__
```
mmichal10: File is now empty, so i believe it can be removed.

robertbaldyga: Done.

src/metadata/metadata_hash.c
```diff
@@ -525,23 +529,28 @@ int ocf_metadata_hash_init(struct ocf_cache *cache,
...
+		core->conf_meta = &core_meta_config[core_id];
+		core->runtime_meta = &core_meta_runtime[core_id];
 	}
 
-	return result;
+	env_spinlock_init(&cache->metadata.lock.eviction);
+	env_rwlock_init(&cache->metadata.lock.status);
+	env_rwsem_init(&cache->metadata.lock.collision);
+
+	return 0;
```
micrakow: why not just return result?

src/ocf_core_priv.h
```diff
@@ -32,12 +32,65 @@ struct ocf_core_io {
...
+	 * on the caching device.
+	 */
+	env_atomic cached_clines;
+	env_atomic dirty_clines;
+	env_atomic initial_dirty_clines;
+
+	env_atomic64 dirty_since;
+
+	struct {
+		/* clines within lru list (?) */
```
micrakow: what about (?) ?

src/metadata/metadata_hash.c
robertbaldyga: To emphasise that when we reach this point, there is no option to fail.

src/ocf_core_priv.h
robertbaldyga: This is genuine copy-paste from `src/ocf_cache_priv.h`. Original comments are preserved.

src/ocf_core_priv.h
```diff
@@ -58,4 +111,7 @@ int ocf_core_volume_type_init(ocf_ctx_t ctx);
 
 void ocf_core_volume_type_deinit(ocf_ctx_t ctx);
 
+#define for_each_core(_cache, _core, _id) \
+	for (_id = 0; _core = &cache->core[_id], _id < OCF_CORE_MAX; _id++)
```
arutk: _core should be updated in each iteration

arutk: Why not iterate over added cores only, like it used to be done?

robertbaldyga: This was meant to be as much generic as possible. We may add specialized version e.g. `for_each_core_added` if we want to.

robertbaldyga: After all I decided to special case iterating over all cores, because it's used much less often and making it default would be error-prone.

PR #182, opened 2019-06-05
Fix order of operations in cache stop completion

mmichal10:
Prevent metadata updater being freed by pyocf before it's deinitialized by ocf.

### DIFF
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -1942,6 +1942,9 @@ static void ocf_mngt_cache_stop_finish(ocf_pipeline_t pipeline,
 	struct ocf_mngt_cache_stop_context *context = priv;
 	ocf_cache_t cache = context->cache;
 	ocf_ctx_t ctx = context->ctx;
+	int pipeline_error;
+	ocf_mngt_cache_stop_end_t pipeline_cmpl;
+	void *completion_priv;
 
 	if (!error) {
 		env_mutex_lock(&ctx->lock);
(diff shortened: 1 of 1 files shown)


src/mngt/ocf_mngt_cache.c
```diff
@@ -1942,6 +1942,9 @@ static void ocf_mngt_cache_stop_finish(ocf_pipeline_t pipeline,
 	struct ocf_mngt_cache_stop_context *context = priv;
 	ocf_cache_t cache = context->cache;
 	ocf_ctx_t ctx = context->ctx;
+	int pipeline_error;
+	ocf_mngt_cache_stop_end_t pipeline_cmpl;
+	void *context_priv;
```
arutk: maybe completion_priv would be a better name?

src/mngt/ocf_mngt_cache.c
```diff
@@ -1972,11 +1975,19 @@ static void ocf_mngt_cache_stop_finish(ocf_pipeline_t pipeline,
 				context->cache_name);
 	}
 
-	context->cmpl(cache, context->priv,
-			error ?: context->cache_write_error);
+	/*
+	 * FIXME This is a temporary workaround to prevent dereferencing pipeline
```
arutk: I propose a different message here:

FIXME: Destroying pipeline before completing management operation is a temporary workaround for insufficient object lifetime management in pyocf. Context must not be referenced after destroying pipeline as this is typically freed upon pipeline destroy.

PR #187, opened 2019-06-10
Create config which excludes tests marked as "long" and security tests by default

KamilLepek:
Signed-off-by: Kamil Lepek <kamil.lepek94@gmail.com>


KamilLepek:
This will enable to execute tests made for CI just by typing "pytest". Security tests and long (nightly) tests will have to be run by specifying the path explicitly.

PR #201, opened 2019-06-19
Fix error handling in case of failed load

mmichal10:
Store default metadata values and restore them in case of failed load.


src/metadata/metadata_hash.c
```diff
@@ -1200,6 +1200,7 @@ struct ocf_metadata_hash_context {
 	void *priv;
 	ocf_pipeline_t pipeline;
 	ocf_cache_t cache;
+	struct ocf_metadata_raw raw_desc[metadata_segment_fixed_size_max];
```
arutk: I suggest renaming 'raw_desc' to something more descriptive like 'segment_copy' or something along those lines.

PR #220, opened 2019-07-24
New hash function formula

arutk:
Preparation for fine-granularity metadata locks implementation:
1. Modifying hash function to return return hashes in increment of 1 (see comment in code)
2. Small cleanup of struct map_info.

### DIFF
--- a/src/metadata/metadata_misc.h
+++ b/src/metadata/metadata_misc.h
@@ -6,21 +6,19 @@
 #ifndef __METADATA_MISC_H__
 #define __METADATA_MISC_H__
 
-/*
- * Hash function needs number that has no common factors with both number
- * of cores and number of entries in metadata hash container (hash lists).
- * This can be easily achived by picking prime which is bigger than maximum
- * number of cores and ensuring that count of hash table entries is not
- * divisible by this number. Let's choose 4099, which is smallest prime
- * greater than OCF_CORE_MAX (which is 4096).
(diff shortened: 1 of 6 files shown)


arutk:
retest this please

src/metadata/metadata_misc.h
```diff
@@ -6,20 +6,16 @@
...
+ * of sorting is required to assure that (future) hash bucket metadata locks are
+ * always acquired in fixed order, eliminating the risk of dead locks.
  */
-#define OCF_HASH_PRIME 4099
-
 static inline ocf_cache_line_t ocf_metadata_hash_func(ocf_cache_t cache,
 		uint64_t core_line_num, ocf_core_id_t core_id)
 {
-	return (ocf_cache_line_t) ((core_line_num * OCF_HASH_PRIME + core_id) %
+	return (ocf_cache_line_t) ((core_line_num  + (core_id << 10U)) %
```
robertbaldyga: In case of big cache and few medium size cores we are going to have more collisions than we had so far. If we want to have basic *increment with overflow* hash function it would be optimal to offset each core by `hash_table_entries / N`, where `N` is **number of cores**, but because number of cores can change dynamically this is hard to achieve. To keep it simple, my best guess is to set N to something between 32 and 128. This minimizes overlaps for small number of cores, but also prevents hitting the same buckets quickly when number of cores increases.

arutk: sounds good

PR #227, opened 2019-08-01
Don't try to remove invalid cores

mmichal10:
If valid cache metadata was read, but environment has changed (i.e. number of cache lines has changed) ocf (in error handling path) was trying to close cores which were not opened. It happened due to cores were marked in cache metadata as added, but any cache inserting operation didn't take place. 

In this patch 'added' flag in cache metadata was replaced with more meaningful 'valid' - it is set if given core is stored in cache metadata. Moreover, new 'added' flag was added to core run-time metadata and it is set if given core is added to cache.

### DIFF
--- a/src/ocf_core_priv.h
+++ b/src/ocf_core_priv.h
@@ -30,8 +30,8 @@ struct ocf_core_meta_config {
 
 	uint8_t type;
 
-	/* This bit means that object was added into cache */
-	uint32_t added : 1;
+	/* This bit means that object was saved in cache metadata */
+	uint32_t valid : 1;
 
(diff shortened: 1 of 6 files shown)


src/ocf_core_priv.h
```diff
@@ -30,8 +30,8 @@ struct ocf_core_meta_config {
 
 	uint8_t type;
 
-	/* This bit means that object was added into cache */
-	uint32_t added : 1;
+	/* This bit means that object was saved in cache metadata */
+	uint32_t in_metadata : 1;
```
arutk: How about naming this "configured" ?

PR #228, opened 2019-08-02
Additional volume type ops: deinit.

mmichal10:
Adapter can't deinitialize volume types before deinitializing core pool.
Additional callback allows ocf to perform this operations in appropriate order.

### DIFF
--- a/src/ocf_ctx.c
+++ b/src/ocf_ctx.c
@@ -209,6 +209,17 @@ void ocf_ctx_get(ocf_ctx_t ctx)
 /*
  *
  */
+static void ocf_ctx_unregister_volume_types(ocf_ctx_t ctx)
+{
+	int id;
+
+	for (id = 0; id < OCF_VOLUME_TYPE_MAX; id++)
+		ocf_ctx_unregister_volume_type(ctx, id);
+}
+
+/*
+ *
+ */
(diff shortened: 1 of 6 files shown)


src/ocf_ctx.c
```diff
@@ -221,7 +232,7 @@ void ocf_ctx_put(ocf_ctx_t ctx)
 	env_rmutex_unlock(&ctx->lock);
 
 	ocf_mngt_core_pool_deinit(ctx);
-	ocf_core_volume_type_deinit(ctx);
+	ocf_ctx_deinit(ctx);
```
robertbaldyga: We gonna have memory leaks, as core_volume type will never be deinitialized.

inc/ocf_ctx.h
```diff
@@ -210,6 +210,9 @@ struct ocf_ctx_ops {
 
 	/* Logger operations */
 	struct ocf_logger_ops logger;
+
+	/* Deinitialize context */
+	void (*deinit)(void);
```
robertbaldyga: Maybe we should just deinitialize all the volume types automatically when context is destroyed, instead of creating callback which only purpose is to perform simple operation which we can easily do internally. This only leads to introduction of boilerplate code in adapters.

PR #237, opened 2019-08-13
pyocf: Fix tests to send io aligned to sector size

robertbaldyga:

### DIFF
--- a/tests/functional/tests/basic/test_pyocf.py
+++ b/tests/functional/tests/basic/test_pyocf.py
@@ -32,8 +32,8 @@ def test_simple_wt_write(pyocf_ctx):
     core_device.reset_stats()
 
     write_data = Data.from_string("This is test data")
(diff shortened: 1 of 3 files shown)


tests/functional/tests/basic/test_pyocf.py
```diff
@@ -32,7 +32,7 @@ def test_simple_wt_write(pyocf_ctx):
     core_device.reset_stats()
 
     write_data = Data.from_string("This is test data")
-    io = core.new_io(cache.get_default_queue(), 20, write_data.size,
+    io = core.new_io(cache.get_default_queue(), 512, write_data.size,
```
arutk: You can use Size.from_sector(1).bytes (not sure about the exact syntax). Would be easier to support configurable sector size in the future.

robertbaldyga: Great suggestion, thanks!

PR #238, opened 2019-08-13
Split cacheline locking implementation into trylock and lock

arutk:
This change refactors the code in order to prepare for removing
global concurrency lock, which won't be needed after per-bucket
metadata locking is in place.

### DIFF
--- a/src/concurrency/ocf_cache_line_concurrency.c
+++ b/src/concurrency/ocf_cache_line_concurrency.c
@@ -665,120 +665,6 @@ static inline void __remove_line_from_waiters_list(struct ocf_cache_line_concurr
 /*
  *
  */
-static int _ocf_req_lock_rd_common(struct ocf_request *req, void *context,
-		__on_lock on_lock)
-{
-	bool locked, waiting;
-	int32_t i;
-	struct ocf_cache_line_concurrency *c = req->cache->device->concurrency.cache_line;
-	ocf_cache_line_t line;
-
-	OCF_DEBUG_RQ(req, "Lock");
-
-	ENV_BUG_ON(env_atomic_read(&req->lock_remaining));
-	ENV_BUG_ON(!on_lock);
-
-	/* Try lock request without adding waiters */
-
-	env_rwlock_read_lock(&c->lock);
-	/* At this point we have many thread that tries get lock for request */
-
-	locked = true;
-	for (i = 0; i < req->core_line_count; i++) {
-
-		if (req->map[i].status == LOOKUP_MISS) {
-			/* MISS nothing to lock */
-			continue;
-		}
-
-		line = req->map[i].coll_idx;
-		ENV_BUG_ON(line >= req->cache->device->collision_table_entries);
-		ENV_BUG_ON(req->map[i].rd_locked);
-		ENV_BUG_ON(req->map[i].wr_locked);
-
-		if (__lock_cache_line_rd(c, line, NULL, NULL, 0)) {
-			/* cache line locked */
-			req->map[i].rd_locked = true;
-		} else {
-			/* Not possible to lock all request */
-			locked = false;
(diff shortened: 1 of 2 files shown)


src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -896,41 +896,48 @@ static int _ocf_req_lock_wr_common(struct ocf_request *req, void *context,
 		ENV_BUG_ON(req->map[i].rd_locked);
 		ENV_BUG_ON(req->map[i].wr_locked);
 
-		if (!__lock_cache_line_wr(c, line, on_lock, context, i)) {
+		if (!__lock_cache_line_wr(c, line, on_lock, req, i)) {
 			/* lock not acquired and not added to wait list */
-			waiting = false;
+			ret = -ENOMEM;
```
imjfckm: Please use OCF error codes

src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -749,90 +771,62 @@ static int _ocf_req_lock_rd_common(struct ocf_request *req, void *context,
 		ENV_BUG_ON(req->map[i].rd_locked);
 		ENV_BUG_ON(req->map[i].wr_locked);
 
-		if (!__lock_cache_line_rd(c, line, on_lock, context, i)) {
+		if (!__lock_cache_line_rd(c, line, on_lock, req, i)) {
 			/* lock not acquired and not added to wait list */
-			waiting = false;
+			ret = -ENOMEM;
```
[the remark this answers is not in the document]
arutk: done

src/concurrency/ocf_cache_line_concurrency.c
arutk: done

src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -720,22 +737,27 @@ static int _ocf_req_lock_rd_common(struct ocf_request *req, void *context,
...
-	env_rwlock_read_unlock(&c->lock);
+	return ret;
+}
 
-	if (locked) {
-		/* Request completely locked, return acquired status */
-		return OCF_LOCK_ACQUIRED;
-	}
+/*
+ *
```
robertbaldyga: Please add comment that function should be called under write lock.

src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -749,90 +771,62 @@ static int _ocf_req_lock_rd_common(struct ocf_request *req, void *context,
...
-			waiting = false;
+			ret = -OCF_ERR_NO_MEM;
 			break;
 		}
 	}
 
-	if (!waiting) {
-		for (; i >= 0; i--)
-			__remove_line_from_waiters_list(c, req, i, context, OCF_READ);
+	if (ret < 0) {
```
robertbaldyga: We can drop one level of indentation by doing:
```
if (ret < 0)
	goto err;
```
and then:
```
err:
	for (; i >= 0; i--)
		__remove_line_from_waiters_list(c, req, i, context, OCF_READ);
	env_atomic_set(&req->lock_remaining, 0);
```

src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -867,22 +861,28 @@ static int _ocf_req_lock_wr_common(struct ocf_request *req, void *context,
...
-	env_rwlock_read_unlock(&c->lock);
+	return ret;
+}
 
-	if (locked) {
-		/* Request completely locked, return acquired status */
-		return OCF_LOCK_ACQUIRED;
-	}
+/*
+ *
```
robertbaldyga: Please add comment that function should be called under write lock.

PR #240, opened 2019-08-14
Documentation early phase

Ostrokrzew:
**doc/**: Added config file and main page's draft for doxygen to document ocf_env.
**env/posix/**: Modify files to prepare ocf_env for documentation in doxygen generator, but this is really early phase. Also done some more cleanup with redundant includes.


Ostrokrzew:
retest this please

tests/unit/tests/ocf_env/ocf_env.h
```diff
@@ -29,7 +29,7 @@
 #include <unistd.h>
 #include <sys/time.h>
 
-#include "ocf_env_list.h"
+#include "../../../../env/posix/ocf_env_list.h"
```
robertbaldyga: I think preferred way would be to use `make env OCF_ENV=posix` (see build test Makefile).
BTW this commit makes few unrelated changes (update documentation headers, comments, partially use posix env in UT, add new doxygen config file, fix python script syntax, cleanup unused includes). Maybe it would be better to split it into few commits each one making one consistent change.

Ostrokrzew: You're right, I will split it. UT cleanup would come with another PR, when I would make this correctly.

PR #243, opened 2019-08-23
Initialize freelist using metadata layout implementation

arutk:
Metadata layout (seq / striping) is already encapsulated
in ocf_metadata_layout_iface, no need to double this logic
in freelist initialization.

### DIFF
--- a/src/metadata/metadata_hash.c
+++ b/src/metadata/metadata_hash.c
@@ -1019,97 +1019,33 @@ static inline void _ocf_init_collision_entry(struct ocf_cache *cache,
 }
 
 /*
- * Default initialization of freelist partition
- */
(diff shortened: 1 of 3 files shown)


src/metadata/metadata_hash.c
```diff
@@ -32,6 +32,8 @@
 
 static void ocf_metadata_hash_init_iface(struct ocf_cache *cache,
 		ocf_metadata_layout_t layout);
+static ocf_cache_line_t ocf_metadata_hash_map_phy2lg_striping(
```
robertbaldyga: This forward declaration is not needed.

arutk: done

PR #244, opened 2019-08-23
Intro to ocf_env documentation and ocf_env unify

Ostrokrzew:
**doc/**: Added config file and main page's draft for doxygen to document ocf_env.
**env/posix/**: Modify files to prepare ocf_env for documentation in doxygen generator, but this is really early phase. Also done some more cleanup with redundant includes.
**add_new_test_file.py**: Added parentheses in 'print' command.
**ocf**: Unused macros cleanup and use of unused but defined macro '__packed'.

### DIFF
--- a/env/posix/ocf_env.h
+++ b/env/posix/ocf_env.h
@@ -7,12 +7,14 @@
  * @file
  * @brief OCF environment functions and macros
  * 
- * This file contains definitions for many OCF functions
- * and sets macros
+ * This file contains definitions for OCF POSIX environment functions and macros.
  */
 
 #ifndef __OCF_ENV_H__
 #define __OCF_ENV_H__
+
+#ifndef __LIBOCF_ENV_H__
+#define __LIBOCF_ENV_H__
 
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
@@ -41,7 +43,8 @@
 
 #include "ocf_env_list.h"
 #include "ocf_env_headers.h"
-#include "ocf/ocf_err.h"
+
+/* *** COMMON MACROS *** */
 
 /* linux sector 512-bytes */
 #define ENV_SECTOR_SHIFT	9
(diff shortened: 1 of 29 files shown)


env/posix/ocf_env.h
```diff
@@ -3,6 +3,14 @@
  * SPDX-License-Identifier: BSD-3-Clause-Clear
  */
 
+/**
+ * @file
+ * @brief OCF environment functions and macros
+ * 
+ * This file contains definitions for many OCF functions
+ * and sets macros
```
arutk: I suggest edit:
This file contains definitions for OCF POSIX environment functions and macros.

env/posix/ocf_env_headers.h
```diff
@@ -3,6 +3,12 @@
  * SPDX-License-Identifier: BSD-3-Clause-Clear
  */
 
+/**
+ * @file
+ * @brief OCF libs and macros
+ *
+ * Header file with OCF common libraries and macros.
```
arutk: Actually looks like this file is mostly to aggregate external header includes. 

Also, where is this file included? I don't see a single reference to it. And it looks like it would make sense to include it from env/posix/ocf_env.h.

Ostrokrzew: It's included in ocf_env.h.

env/posix/ocf_env.h
Ostrokrzew: Changed as You suggested.

tests/unit/tests/ocf_env/ocf_env_headers.h
```diff
@@ -1,13 +1,28 @@
...
 #ifndef __OCF_ENV_HEADERS_H__
 #define __OCF_ENV_HEADERS_H__
 
 #include <stdint.h>
-#include <stdbool.h>
 #include <stddef.h>
+#include <stdbool.h>
+
+/* TODO: Move prefix printing to context logger. */
+#define OCF_LOGO "Intel(R) OCF"
```
arutk: We shouldn't use "Intel" branding anywhere in Open CAS * projects

Ostrokrzew: You're right, thanks.

PR #247, opened 2019-08-28
 Per-execution-context freelists

arutk:
Global free cacheline list is divided into a set of freelists, one
per execution context. When attempting to map addres to cache, first
the freelist for current execution context is considered (fast path).
If current execution context freelist is empty (fast path failure),
mapping function attempts to get freelist from other execution context
list (slow path).

The purpose of this change is improve concurrency in freelist access.
It is part of fine granularity metadata lock implementation.


env/posix/ocf_env.c
```diff
@@ -130,8 +131,57 @@ void env_stack_trace(void)
...
+pthread_mutex_t exec_context_list_mutex = PTHREAD_MUTEX_INITIALIZER;
+pthread_mutex_t *exec_context_mutex = NULL;
+
+static void init_execution_context(void)
+{
+	unsigned count = env_get_execution_context_count();
+	unsigned i;
+
+	ENV_BUG_ON(count == -1);
+	exec_context_mutex = malloc(count * sizeof(exec_context_mutex[0]));
```
robertbaldyga: Memory leak: this memory is never freed. Also mutexes created below are never deinitialized.
Maybe we should introduce `env_init()`/`env_deinit()` functions that would called before first and after last env_ call? Then we would call them during intialization/deinitialization of ocf_ctx (with some kind of refcount). Or even this refcount may be moved to env, and then we would have `env_get()` that initializes env if it's not initialized yet and increments refcount, and `env_put()` that decrements refcount and deinitializes env when refcount reaches zero.

env/posix/ocf_env.c
```diff
@@ -130,8 +131,57 @@ void env_stack_trace(void)
...
+
+	pthread_mutex_lock(&exec_context_list_mutex);
+	if (exec_context_mutex == NULL)
+		init_execution_context();
+	pthread_mutex_unlock(&exec_context_list_mutex);
+
+	cpu = sched_getcpu();
+	cpu = (cpu == -1) ?  0 : cpu;
+
+	ENV_BUG_ON(pthread_mutex_lock(exec_context_mutex + cpu));
```
robertbaldyga: Using index operator instead of doing explicit pointer arithmetic would look much clearer.

env/posix/ocf_env.c
```diff
@@ -130,8 +131,57 @@ void env_stack_trace(void)
...
+
+	return cpu;
+}
+
+void env_put_execution_context(unsigned ctx)
+{
+	pthread_mutex_unlock(exec_context_mutex + ctx);
```
robertbaldyga: Ditto.

env/posix/ocf_env.c
imjfckm: I think that moving env from stateless to stateful is a step in dangerous direction. What about ` __attribute__((destructor)) `? This way state specific to POSIX env is destroyed in POSIX specific way, at shared object unload or program end.

arutk: Went with GCC constructor/destructor attributes to assure proper cleanup and initialization.

src/ocf_freelist.c
```diff
@@ -0,0 +1,381 @@
...
+
+	*cline = freelist->part[ctx].head;
+	_ocf_freelist_remove_cache_line(freelist, ctx, *cline);
+
+	ocf_freelist_unlock(freelist, ctx);
+
+	return true;
+}
+
+static int get_next_freelist(ocf_freelist_t freelist)
```
arutk: consider having independent victim idx per exec ctx

env/posix/ocf_env.c
```diff
@@ -130,8 +131,64 @@ void env_stack_trace(void)
...
+ * result the caller might actually get preempted, but no other thread will
+ * execute in this context by the time the caller puts current execution ctx. */
+unsigned env_get_execution_context(void)
+{
+	unsigned cpu;
+
+	cpu = sched_getcpu();
+	cpu = (cpu == -1) ?  0 : cpu;
+
+	ENV_BUG_ON(pthread_mutex_lock(exec_context_mutex + cpu));
```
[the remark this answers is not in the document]
arutk: done

env/posix/ocf_env.c
```diff
@@ -130,8 +131,64 @@ void env_stack_trace(void)
...
+
+	return cpu;
+}
+
+void env_put_execution_context(unsigned ctx)
+{
+	pthread_mutex_unlock(exec_context_mutex + ctx);
```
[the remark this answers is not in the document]
arutk: done

src/ocf_freelist.c
arutk: This doesn't seem to improve things much. To make this work we would need to be pinned to current execution context for the duration of slow path. Current implementation only pins execution to exec context for fast path.

PR #252, opened 2019-08-29
Update OCF README.md

felipe-barajas:
The doxygen API documentation link was being referenced twice. The first link should have pointed to https://open-cas.github.io/getting_started_ocf.html instead.


KamilLepek:
ok to test

PR #254, opened 2019-09-05
Unify ocf_env for regular use and unit tests

Ostrokrzew:
**ocf_env**: unified and deleted separate files for regular use and UT,
now UT uses the same files as 'regular' OCF;
	one of macros (env_memcpy) is switched on parallel function when UT
runs, due to compilation issues;
	changed 'Intel OCF' on 'OCF' due to earlier suggestion.
**prepare script**: added method to set 'TEST' flag when OCF is used by UT;
	added missing parentheses in one of 'print' functions;
	changed omitted phrase 'python3.6' on 'python3';
	removing 'static inline' keywords from UT's copy of ocf_env.h file,
due to compilation issues.
**run tests script**: added cleanup function on startup,
before actual start of UT.
**alru cleaning policy**: little update in functions to properly use atomic
values after unify of ocf_env.
**general**: fixed typo 'warp/ wrap';
	deleted recursive inclusions;
	some aesthetic changes in comments.


tests/unit/framework/run_unit_tests.py
```diff
@@ -19,10 +19,30 @@ def run_command(args):
...
 
 
+def cleanup():
+    repo_dir = os.path.join(tests_config.MAIN_DIRECTORY_OF_TESTED_PROJECT)
+    script_path = os.path.dirname(os.path.realpath(__file__))
+    result = run_command(["cd", repo_dir])
+    if result.returncode != 0:
+        raise Exception("Cleanup before testing failed!")
+
+    result = run_command(["git clean", "-dxf"])
```
KamilLepek: Git clean seems to be too much, e.g. when someone has some changes that are not staged and wants to run unit tests he will get his changes wiped out.

Ostrokrzew: Fixed

PR #261, opened 2019-09-13
Fixed some bugs found by the coverity tool

micrakow:
Signed-off-by: Michal Rakowski <michal.rakowski@intel.com>

### DIFF
--- a/src/metadata/metadata_raw.c
+++ b/src/metadata/metadata_raw.c
@@ -373,7 +373,7 @@ static int _raw_ram_flush_do_asynch_fill(ocf_cache_t cache,
 	uint32_t raw_page;
 	struct _raw_ram_flush_ctx *ctx = context;
 	struct ocf_metadata_raw *raw = NULL;
-	uint64_t size;
+	uint32_t size;
 
 	ENV_BUG_ON(!ctx);
 
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -571,9 +571,14 @@ static int _ocf_mngt_init_new_cache(struct ocf_cache_mngt_init_params *params)
 		return -OCF_ERR_NO_MEM;
 	}
 
+	if (!ocf_refcnt_inc(&cache->refcnt.cache)){
+		env_mutex_destroy(&cache->flush_mutex);
+		env_vfree(cache);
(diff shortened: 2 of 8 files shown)


src/metadata/metadata_raw.c
```diff
@@ -380,7 +380,7 @@ static int _raw_ram_flush_do_asynch_fill(ocf_cache_t cache,
 	raw = ctx->raw;
 	ENV_BUG_ON(!raw);
 
-	size = raw->entry_size * raw->entries_in_page;
+	size = (uint64_t) raw->entry_size * raw->entries_in_page;
```
robertbaldyga: `size` doesn't need to be 64-bit long. In fact total size of entries within page can never be bigger than 4096 bytes, so even `uint16_t` should be more than enough.

src/mngt/ocf_mngt_cache.c
```diff
@@ -1654,14 +1659,8 @@ int ocf_mngt_cache_start(ocf_ctx_t ctx, ocf_cache_t *cache,
...
-	} else {
-		if (cfg->name) {
-			ocf_log(ctx, log_err, "Inserting cache %s failed\n",
-					cfg->name);
-		} else {
-			ocf_log(ctx, log_err, "Inserting cache failed\n");
-		}
-	}
+	} else
+		ocf_log(ctx, log_err, "Inserting cache failed\n");
```
robertbaldyga: As cache name is mandatory now, I'd rather leave the variant with name, which is more informative.

src/mngt/ocf_mngt_io_class.c
```diff
@@ -295,8 +295,8 @@ int ocf_mngt_cache_io_classes_configure(ocf_cache_t cache,
 
 out_edit:
 	if (result) {
-		ENV_BUG_ON(env_memcpy(cache->user_parts, sizeof(&cache->user_parts),
-					old_config, sizeof(&cache->user_parts)));
+		ENV_BUG_ON(env_memcpy(cache->user_parts, sizeof(cache->user_parts),
```
robertbaldyga: I believe we have similar issue in line 280.

micrakow:
retest this please

micrakow:
retest this please

micrakow:
retest this please

PR #270, opened 2019-09-18
Pp change param api

imjfckm:


src/promotion/nhit/nhit.c
```diff
@@ -110,7 +110,7 @@ ocf_error_t nhit_set_param(ocf_promotion_policy_t policy, uint8_t param_id,
 }
 
 ocf_error_t nhit_get_param(ocf_promotion_policy_t policy, uint8_t param_id,
-		uint64_t *param_value)
+		uint32_t *param_value)
```
arutk: Wouldn't it be cleaner if you stored the trigger threshold param in [%] as 32 bit integer, and have a separate 64 bit atomic with threshold value in [clines] to use in I/O path? 

Are you sure the current implementation would always return the same trigger threshold % as originally set? I'm concerned with potential rounding errors in get/set calculations. What I propose would alleviate  this kind of concerns. Also there would be no doubt that 32 bit integer is wide enough to store the threshold value.

imjfckm: IMO it is better just to return the value we're using for I/O path. There is a slight possibility that you set 80% and get 81% back after calculations and rounding, but at least we're reporting the value we're using internally.

Also, I don't think there is any possibility to overflow since we're performing symmetric operations.

PR #283, opened 2019-09-23
Accept max values of nhit PP as valid.

rafalste:
Signed-off-by: Rafal Stefanowski <rafal.stefanowski@intel.com>


arutk:
ok to test

PR #284, opened 2019-09-24
Prevent cache name duplicate

mmichal10:


src/ocf_cache.c
```diff
@@ -20,7 +20,17 @@ ocf_volume_t ocf_cache_get_volume(ocf_cache_t cache)
 
 int ocf_cache_set_name(ocf_cache_t cache, const char *src, size_t src_size)
 {
+	ocf_cache_t tmp_cache;
+	int result;
 	OCF_CHECK_NULL(cache);
+
+	result = ocf_mngt_cache_get_by_name(ocf_cache_get_ctx(cache), src,
```
robertbaldyga: This function was meant to be simple setter. It would be better to have this check in _ocf_mngt_cache_start() before this function is called.

mmichal10: Since it is a part of public API, shouldn't we try to prevent setting invalid cache name?

src/mngt/ocf_mngt_cache.c
```diff
@@ -1201,10 +1205,16 @@ static int _ocf_mngt_cache_start(ocf_ctx_t ctx, ocf_cache_t *cache,
...
 
+	result = env_rmutex_lock_interruptible(&ctx->lock);
+	if (result)
+		goto _cache_mngt_init_instance_ERROR;
+
 	/* Prepare cache */
 	result = _ocf_mngt_init_prepare_cache(&params, cfg);
-	if (result)
+	if (result) {
+		env_rmutex_unlock(&ctx->lock);
```
micrakow: how about moving that to some goto-label?

inc/ocf_err.h
```diff
@@ -116,6 +116,9 @@ typedef enum {
 
 	/** Invalid cache line size */
 	OCF_ERR_INVALID_CACHE_LINE_SIZE,
+
+	/** Invalid cache name loaded*/
+	OCF_ERR_CACHE_NAME_MISMATCH,
```
micrakow: Please update PyOcf as well

src/ocf_cache.c
robertbaldyga: Oops, it looks like a oversight. This function definitely should not be public.

src/mngt/ocf_mngt_cache.c
```diff
@@ -547,33 +558,29 @@ static void _ocf_mngt_init_instance_load(
...
 	ocf_cache_t cache = env_vzalloc(sizeof(*cache));
+	int result;
 
 	if (!cache)
 		return -OCF_ERR_NO_MEM;
 
 	if (ocf_mngt_cache_lock_init(cache)) {
-		env_vfree(cache);
-		return -OCF_ERR_NO_MEM;
+		result = -OCF_ERR_NO_MEM;
```
arutk: Now we will attempt to deinit uninitialized lock in this path. You need to add a separate label to handle this correctly.

PR #290, opened 2019-09-25
Add name_len field to config structures taking cache/core name

Ostrokrzew:


src/mngt/ocf_mngt_core.c
```diff
@@ -541,8 +541,8 @@ void ocf_mngt_cache_add_core(ocf_cache_t cache,
 		goto err_pipeline;
 	}
 
-	result = env_strncpy(name, OCF_CORE_NAME_SIZE,
-			cfg->name, OCF_CORE_NAME_SIZE);
+	result = env_strncpy(name, cfg->name_len,
```
arutk: second param should be OCF_CORE_NAME_SIZE

inc/ocf_mngt.h
```diff
@@ -237,6 +243,11 @@ struct ocf_mngt_cache_config {
 	 */
 	char name[OCF_CACHE_NAME_SIZE];
 
+	/**
+	 * @brief Cache name length
```
arutk: Maybe let's use the comment from strncpy_s:
@brief The maximum number of characters to copy from name

inc/ocf_mngt.h
```diff
@@ -60,6 +65,7 @@ struct ocf_mngt_core_config {
 static inline void ocf_mngt_core_config_set_default(
 		struct ocf_mngt_core_config *cfg)
 {
+	cfg->name_len = sizeof(*cfg->name);
```
arutk: This is wrong. name_len is supposed to store size of the string. Please remove this from set_default function - whoever sets cache name is responsible for setting this value as well.

inc/ocf_mngt.h
```diff
@@ -302,6 +313,7 @@ struct ocf_mngt_cache_config {
 static inline void ocf_mngt_cache_config_set_default(
 		struct ocf_mngt_cache_config *cfg)
 {
+	cfg->name_len = sizeof(*cfg->name);
```
arutk: This is wrong. name_len is supposed to store size of the string. Please remove this from set_default function - whoever sets cache name is responsible for setting this value as well.

inc/ocf_mngt.h
arutk: same comment applies to core config

src/mngt/ocf_mngt_core.c
Ostrokrzew: done

example/simple/src/main.c
```diff
@@ -91,7 +91,7 @@ static void simple_complete(ocf_cache_t cache, void *priv, int error)
  */
 int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
 {
-	struct ocf_mngt_cache_config cache_cfg = { .name = "cache1" };
+	struct ocf_mngt_cache_config cache_cfg = { .name = "cache1", .name_len = env_strnlen(cache_cfg.name, OCF_CACHE_NAME_SIZE) };
```
arutk: name_len should be assigned "env_strnlen(...) + 1".   Or even better" sizeof("cache1")

example/simple/src/main.c
```diff
@@ -208,6 +208,7 @@ int initialize_core(ocf_cache_t cache, ocf_core_t *core)
 	/* Core configuration */
 	ocf_mngt_core_config_set_default(&core_cfg);
 	core_cfg.name = "core1";
+	core_cfg.name_len = env_strnlen(core_cfg.name, OCF_CORE_NAME_SIZE);
```
arutk: ... same here - name_len should be set to the size of string memory buffer, not the actual string length. Let's discuss live

src/mngt/ocf_mngt_core.c
arutk: I meant second out of four, not the second one that you change :)

env_strncpy(name, OCF_CORE_NAME_SIZE, cfg->name, cfg->name_len)

example/simple/src/main.c
```diff
@@ -91,7 +91,8 @@ static void simple_complete(ocf_cache_t cache, void *priv, int error)
  */
 int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
 {
-	struct ocf_mngt_cache_config cache_cfg = { .name = "cache1" };
+	struct ocf_mngt_cache_config cache_cfg = { .name = "cache1",
+											   .name_len = sizeof("cache1") };
```
Ostrokrzew: in editor this line starts under '.name'

src/mngt/ocf_mngt_core.c
Ostrokrzew: my mistake, now fixed

example/simple/src/main.c
Ostrokrzew: changed

example/simple/src/main.c
Ostrokrzew: fixed and looking for discussion

example/simple/src/main.c
robertbaldyga: So probably you have tab length set to 4 chars instead of 8.

Ostrokrzew: that was true

PR #295, opened 2019-09-27
tests: functional: Add pretty random generators

### LINKED ISSUE #277: Random numbers generators in security tests are inefficient.
They allocate the numbers instead of creating a generator. This might be the cause of a serious bottleneck while running full security tests suite. We would also like to have the ability to exclude certain ranges from the generator.

robertbaldyga:
Fixes #277 


tests/functional/tests/security/conftest.py
```diff
@@ -11,33 +11,33 @@
...
 
 
-@pytest.fixture(params=get_random_ints(c_uint64))
+@pytest.fixture(params=RandomGenerator(DefaultRanges.UINT64))
 def c_uint64_randomize(request):
     return request.param
 
 
-@pytest.fixture(params=get_random_ints(c_int))
+@pytest.fixture(params=RandomGenerator(DefaultRanges.INT))
```
katlapinka: I think it could be params=RandomGenerator(), because DefaultRanges INT by default

tests/functional/tests/utils/random.py
```diff
@@ -0,0 +1,90 @@
...
+                      string.hexdigits]:
+                yield ''.join(random.choice(t) for _ in range(
+                    random.randint(len_range.min, len_range.max)
+                ))
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        if self.n > self.count:
```
arutk: should be >= to terminate iteration after exactly self.count elements

tests/functional/tests/utils/random.py
```diff
@@ -0,0 +1,90 @@
...
+        self.min = min_val
+        self.max = max_val
+
+    def is_within(self, val):
+        return val >= self.min and val <= self.max
+
+
+class DefaultRanges(Range, enum.Enum):
+    UINT8 = 0, c_uint8(-1).value
+    UINT16 = 0, c_uint8(-1).value
```
arutk: shouldn't this be c_uint16?

tests/functional/tests/utils/random.py
robertbaldyga: Done.

tests/functional/tests/utils/random.py
robertbaldyga: Done.

PR #298, opened 2019-09-30
Core name - pointer to array change

Ostrokrzew:
**ocf_mngt.h**: In core name change
pointer type to array which is 32 characters long;
**core.py**: Add missing import and modify class' field type
to keep consistency;
**ocf_mngt_core**: Remove local variable 'name';
remove env_vmalloc for 'name' - isn't no longer needed;
remove initialization 'name' - as above;
remove env_vfree for context->cfg.name - variable isn't no allocated
in memory;
check if cfg->name exists;
change label in goto from deleted err_name to the closest err_pipeline.

### DIFF
--- a/src/mngt/ocf_mngt_core.c
+++ b/src/mngt/ocf_mngt_core.c
@@ -476,7 +476,6 @@ static void ocf_mngt_cache_add_core_finish(ocf_pipeline_t pipeline,
 
 out:
 	context->cmpl(cache, core, context->priv, error);
-	env_vfree(context->cfg.name);
 	env_vfree(context->cfg.uuid.data);
 	ocf_pipeline_destroy(context->pipeline);
 }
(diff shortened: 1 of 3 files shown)


Ostrokrzew:
retest this please

src/mngt/ocf_mngt_core.c
```diff
@@ -535,23 +533,10 @@ void ocf_mngt_cache_add_core(ocf_cache_t cache,
 	context->cache = cache;
 	context->cfg = *cfg;
 
-	name = env_vmalloc(OCF_CORE_NAME_SIZE);
```
micrakow: check at line 521 is no longer needed as well :)

Ostrokrzew: thanks for caution

micrakow: *should be changed as we discussed

PR #312, opened 2019-10-03
Rand generator improvements

robertbaldyga:


tests/functional/tests/security/test_management_start_fuzzy.py
```diff
@@ -21,38 +21,30 @@ def try_start_cache(**config):
     cache = Cache.start_on_device(cache_device, **config)
     cache.stop()
 
-
 @pytest.mark.security
 @pytest.mark.parametrize("cls", CacheLineSize)
-def test_fuzzy_start_cache_mode(pyocf_ctx, cls, c_uint32_randomize):
+def test_fuzzy_start_cache_mode(pyocf_ctx, cls, not_cache_mode_randomize):
```
mmichal10: Since we don't have tests (yet) for valid parameters values I believe we shouldn't exclude them from range. As we saw yesterday, they can detect bugs. How about applying your change after we extend out test suit with tests for valid values?

PR #313, opened 2019-10-08
Fix cache stats updating.

mmichal10:
When single request to cache was issued, stats updating function was called with
0 bytes as value to update. In case of many request issued to cache, stats were
updated only in case of error.


mmichal10:
retest this please

PR #320, opened 2019-12-05
Limit number of concurrent io submitted by metadata_io_i_asynch()

### LINKED ISSUE #318: OCF should throttle queue depth during flushing metadata
In current implementation of metadata_io_i_asynch function there is no mechanism to throttle queue depth. This mechanism is needed in environments based on model without interruptions - like SPDK, now in the worth case scenario ocf can double memory consumption (can consume double size of metadata stored on cache device) during flushing metadata to cache device.

In case when we have enough memory to handle situation described above, starting cache still can fail due to io submit fail caused by running out of space in submission queue.

robertbaldyga:
Fixes #318

### DIFF
--- a/src/metadata/metadata_io.c
+++ b/src/metadata/metadata_io.c
@@ -177,130 +177,127 @@ int metadata_io_read_i_atomic(ocf_cache_t cache, ocf_queue_t queue, void *priv,
 	return 0;
 }
 
-static void metadata_io_i_asynch_end(struct metadata_io_request *request,
-		int error);
-
-static int ocf_restart_meta_io(struct ocf_request *req);
-
-static struct ocf_io_if meta_restart_if = {
-		.read = ocf_restart_meta_io,
-		.write = ocf_restart_meta_io
-};
-
-static void metadata_io_i_asynch_cmpl(struct ocf_io *io, int error)
+static void metadata_io_req_fill(struct metadata_io_request *m_req)
 {
-	struct metadata_io_request *request = io->priv1;
+	ocf_cache_t cache = m_req->cache;
+	int i;
 
-	metadata_io_i_asynch_end(request, error);
+	for (i = 0; i < m_req->count; i++) {
+		m_req->on_meta_fill(cache, m_req->data,
+			m_req->page + i, m_req->context);
+	}
+}
 
+static void metadata_io_req_drain(struct metadata_io_request *m_req)
+{
+	ocf_cache_t cache = m_req->cache;
+	int i;
+
+	for (i = 0; i < m_req->count; i++) {
+		m_req->on_meta_drain(cache, m_req->data,
+			m_req->page + i, m_req->context);
+	}
+}
+
+static void metadata_io_io_end(struct metadata_io_request *m_req, int error);
+
+static void metadata_io_io_cmpl(struct ocf_io *io, int error)
+{
+	metadata_io_io_end(io->priv1, error);
 	ocf_io_put(io);
 }
 
-static void metadata_io_req_fill(struct metadata_io_request *meta_io_req)
+static int metadata_io_restart_req(struct ocf_request *req)
 {
-	ocf_cache_t cache = meta_io_req->cache;
-	int i;
-
-	for (i = 0; i < meta_io_req->count; i++) {
-		meta_io_req->on_meta_fill(cache, meta_io_req->data,
-			meta_io_req->page + i, meta_io_req->context);
-	}
-}
-
-static void metadata_io_req_drain(struct metadata_io_request *meta_io_req)
-{
-	ocf_cache_t cache = meta_io_req->cache;
-	int i;
-
-	for (i = 0; i < meta_io_req->count; i++) {
-		meta_io_req->on_meta_drain(cache, meta_io_req->data,
-			meta_io_req->page + i, meta_io_req->context);
-	}
-}
-
-static int ocf_restart_meta_io(struct ocf_request *req)
-{
-	struct metadata_io_request *meta_io_req = req->priv;
+	struct metadata_io_request *m_req = req->priv;
 	ocf_cache_t cache = req->cache;
 	struct ocf_io *io;
 	int ret;
 
 	/* Fill with the latest metadata. */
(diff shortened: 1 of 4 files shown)


src/metadata/metadata_io.c
```diff
@@ -218,106 +199,128 @@ static void metadata_io_req_drain(struct metadata_io_request *meta_io_req)
...
+	ctx_data_free(cache->owner, m_req->data);
 
-	/*
-	 * If it's last request, we mark is as finished
-	 * after calling IO end callback
-	 */
-	env_atomic_set(&request->finished, 1);
-	ocf_metadata_updater_kick(cache);
+	if (env_atomic_dec_return(&a_req->req_remaining) == 0) {
+		a_req->on_complete(cache, a_req->context, a_req->error);
```
arutk: It should be possible to move a_req->on_complete call to metadata_io_io_end. This way request will complete sooner. Deallocation would still be done here in metadata_io_req_end after request is removed from updater in_progress list.

src/metadata/metadata_io.c
```diff
@@ -345,140 +343,54 @@ static int metadata_io_i_asynch(ocf_cache_t cache, ocf_queue_t queue, int dir,
...
-		ocf_cache_log(cache, log_warn,
-				"No memory during metadata IO\n");
-		return -OCF_ERR_NO_MEM;
-	}
 	/* IO Requests initialization */
-	for (i = 0; i < io_count; i++) {
-		env_atomic_set(&(a_req->reqs[i].req_remaining), 1);
-		env_atomic_set(&(a_req->reqs[i].finished), 0);
-		a_req->reqs[i].asynch = a_req;
+	for (i = 0; i < ARRAY_SIZE(a_req->reqs); i++) {
```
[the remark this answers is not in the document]
robertbaldyga: Fixed.

src/metadata/metadata_io.c
```diff
@@ -218,106 +199,128 @@ static void metadata_io_req_drain(struct metadata_io_request *meta_io_req)
...
+static void metadata_io_req_submit(struct metadata_io_request *m_req)
 {
-	return ocf_volume_get_max_io_size(&cache->device->volume) / PAGE_SIZE;
+	struct metadata_io_request_asynch *a_req = m_req->asynch;
+	uint32_t max_count = metadata_io_max_page(m_req->cache);
+	uint32_t curr;
+
+	curr = env_atomic_inc_return(&a_req->req_current);
+
+	if (curr * max_count >= a_req->count) {
```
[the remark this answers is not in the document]
robertbaldyga: Fixed.

src/metadata/metadata_io.c
robertbaldyga: Good point. Done.

src/metadata/metadata_io.c
```diff
@@ -177,147 +177,184 @@ int metadata_io_read_i_atomic(ocf_cache_t cache, ocf_queue_t queue, void *priv,
...
+}
 
-	if (env_atomic_dec_return(&a_req->req_remaining)) {
-		env_atomic_set(&request->finished, 1);
-		ocf_metadata_updater_kick(cache);
-		return;
-	}
+static void metadata_io_req_submit(struct metadata_io_request *m_req)
+{
+	env_atomic_set(&(m_req->finished), 0);
```
arutk: We never add brackets after &

src/metadata/metadata_io.c
```diff
@@ -345,140 +379,54 @@ static int metadata_io_i_asynch(ocf_cache_t cache, ocf_queue_t queue, int dir,
...
+		m_req->on_meta_drain = io_hndl;
+		m_req->req.io_if = &metadata_io_restart_if;
+		m_req->req.io_queue = queue;
+		m_req->req.cache = cache;
+		m_req->req.priv = m_req;
+		m_req->req.info.internal = true;
+		m_req->req.rw = dir;
+		m_req->req.map = LIST_POISON1;
+
+		m_req->data = ctx_data_alloc(cache->owner, max_count);
```
arutk: Please change this so that we allocate only 'count' pages here if 'count < max_count' (the case where only one slave request is allocated). Adding extra variable would do:

unsigned data_count = min(count, max_count);

src/metadata/metadata_io.c
```diff
@@ -177,147 +177,184 @@ int metadata_io_read_i_atomic(ocf_cache_t cache, ocf_queue_t queue, void *priv,
...
+	uint32_t curr;
+
+	if (a_req->error) {
+		metadata_io_req_end(m_req);
+		return;
+	}
+
+	curr = env_atomic_inc_return(&a_req->req_current);
+
+	if (curr >= OCF_DIV_ROUND_UP(a_req->count, max_count)) {
```
arutk: I think this should be '>':
if (curr > OCF_DIV_ROUND_UP(a_req->count, max_count)) {

robertbaldyga: The `curr` is counted from 0, so we maximum `curr` should be io count - 1.

src/metadata/metadata_io.c
robertbaldyga: Done.

src/metadata/metadata_io.c
robertbaldyga: Done.

src/metadata/metadata_io.c
```diff
@@ -345,140 +379,59 @@ static int metadata_io_i_asynch(ocf_cache_t cache, ocf_queue_t queue, int dir,
...
+		m_req->on_meta_drain = io_hndl;
+		m_req->req.io_if = &metadata_io_restart_if;
+		m_req->req.io_queue = queue;
+		m_req->req.cache = cache;
+		m_req->req.priv = m_req;
+		m_req->req.info.internal = true;
+		m_req->req.rw = dir;
+		m_req->req.map = LIST_POISON1;
+
+		/* If req_count == io_count and count is not muliple of
```
arutk: *multiple

robertbaldyga: Done.

PR #321, opened 2019-12-09
Fix counting occupancy on WB write insert error

imjfckm:
During error handling in WB write insert we didn't invalidate affected
cache lines. Because of that the cache stopped properly (as it's
supposed to), but cache lines were marked as inserted which caused
occupancy stats to increase even though nothing was succesfully
inserted.

### DIFF
--- a/src/engine/engine_wb.c
+++ b/src/engine/engine_wb.c
@@ -8,6 +8,7 @@
 #include "cache_engine.h"
 #include "engine_common.h"
 #include "engine_wb.h"
+#include "engine_inv.h"
 #include "../metadata/metadata.h"
 #include "../ocf_request.h"
 #include "../utils/utils_io.h"
@@ -98,11 +99,9 @@ static void _ocf_write_wb_complete(struct ocf_request *req, int error)
 	if (req->error) {
 		ocf_engine_error(req, true, "Failed to write data to cache");
 
-		ocf_req_unlock_wr(req);
-
 		req->complete(req, req->error);
 
-		ocf_req_put(req);
+		ocf_engine_invalidate(req);
 	} else {
 		ocf_engine_push_req_front_if(req, &_io_if_wb_flush_metadata,
 				true);


src/engine/engine_wb.c
```diff
@@ -98,11 +99,9 @@ static void _ocf_write_wb_complete(struct ocf_request *req, int error)
 	if (req->error) {
 		ocf_engine_error(req, true, "Failed to write data to cache");
 
-		ocf_req_unlock_wr(req);
-
 		req->complete(req, req->error);
 
-		ocf_req_put(req);
+		ocf_engine_invalidate(req);
```
arutk: Is this really the expected behavior? What if the cachelines were already inserted before the write started? One can argue that since the write was issued, the target LBAs are actually hot and should be inserted to cache. And since we're returning error here, the actual data in these cachelines is irrelevant.

arutk: After giving this some more though I agree that invalidating would probably be fine here, since after getting data write error on a SSD drive we would probably be unable to update metadata, causing cache emergency stop etc.

arutk:
@michalwy please review

src/engine/engine_wb.c
robertbaldyga: @arutk This may be true for most of the SSDs, but in general we cannot assume permanent disk unavailability after single failed io. Especialy in the era of NVMeOF.

arutk: @robertbaldyga what is your recommendation? Are you against this change?

arutk: One more consideration. If we decide not to invalidate these cachelines, subsequent read would potentially allow user to read uninitialized cache device page (if the failed write inserted cache lines and did not modify on-disk pages). This might potentially be a security concern.

michalwy: At the point of added ocf_engine_invalidate() call, cache is already stopped - ocf_engine_error() with stop_cache = true argument is called few lines above ocf_engine_invalidate(). We are already in emergency state at this point, so invalidating seems to be fine here.

PR #326, opened 2019-12-17
Change error code

### LINKED ISSUE #317: OCF should return appropriate error code in case of out of memory
Now, in case of allocation error during start cache ocf returns OCF_ERR_START_CACHE_FAIL instead of ENOMEM. There is no also message about allocation error so user can be confused.

Ostrokrzew:
Change 'OCF_ERR_START_CACHE_FAIL' to 'OCF_ERR_NO_MEM' while CAS fails in case of memory lack on device.
closes #317 issue

Signed-off-by: Ostrokrzew <slawomir.jankowski@intel.com>

### DIFF
--- a/inc/ocf_err.h
+++ b/inc/ocf_err.h
@@ -117,8 +117,11 @@ typedef enum {
 	/** Invalid cache line size */
 	OCF_ERR_INVALID_CACHE_LINE_SIZE,
 
-	/** Invalid cache name loaded*/
+	/** Invalid cache name loaded */
 	OCF_ERR_CACHE_NAME_MISMATCH,
+
+	/** Device does not meet requirements */
+	OCF_ERR_INVAL_CACHE_DEV,
 } ocf_error_t;
 
 #endif /* __OCF_ERR_H__ */
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -635,7 +635,7 @@ static void _ocf_mngt_attach_cache_device(ocf_pipeline_t pipeline,
 	if (context->volume_size < OCF_CACHE_SIZE_MIN) {
 		ocf_cache_log(cache, log_err, "ERROR: Cache cache size must "
 			"be at least %llu [MiB]\n", OCF_CACHE_SIZE_MIN / MiB);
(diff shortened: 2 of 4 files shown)


src/mngt/ocf_mngt_cache.c
```diff
@@ -635,7 +635,7 @@ static void _ocf_mngt_attach_cache_device(ocf_pipeline_t pipeline,
 	if (context->volume_size < OCF_CACHE_SIZE_MIN) {
 		ocf_cache_log(cache, log_err, "ERROR: Cache cache size must "
 			"be at least %llu [MiB]\n", OCF_CACHE_SIZE_MIN / MiB);
-		OCF_PL_FINISH_RET(context->pipeline, -OCF_ERR_START_CACHE_FAIL);
+		OCF_PL_FINISH_RET(context->pipeline, -OCF_ERR_NO_MEM);
```
micrakow: Is this error related to the 'low memory' issue or is it related to the too small device?

robertbaldyga: I agree with @micrakow, this is not memory related problem. Currently we don't have an error code addressing this particular issue, so you have two options:
a) Act as if one of parameters (cache device name) was invalid, because it represents device that doesn't satisfy our requirements, thus return not so very informative `-OCF_ERR_INVAL`.
b) Add dedicated error code and return it.
The latter may be used in `casadm` to print more detailed error message and thus improve UX, so I'd opt for that one.

Ostrokrzew: Hope it's now fixed.

inc/ocf_err.h
```diff
@@ -117,8 +117,11 @@ typedef enum {
 	/** Invalid cache line size */
 	OCF_ERR_INVALID_CACHE_LINE_SIZE,
 
-	/** Invalid cache name loaded*/
+	/** Invalid cache name loaded */
 	OCF_ERR_CACHE_NAME_MISMATCH,
+
+	/** Device does not meet requirements */
+	OCF_ERR_INVAL_DEV_PARAMS,
```
robertbaldyga: How about `OCF_ERR_INVAL_CACHE_DEV`?

KamilLepek:
retest this please

PR #332, opened 2019-12-23
Keep stop pipeline in struct cache.

mmichal10:
To eliminate possibility of allocation error in cache stop, pipeline is
allocated on attach.

Due this change, stop won't return any error, so adapter don't need to care
about rollbacks.

### DIFF
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -1153,6 +1153,8 @@ static void _ocf_mngt_attach_handle_error(
 
 	if (context->flags.device_alloc)
 		env_vfree(cache->device);
+
+	ocf_pipeline_destroy(cache->stop_pipeline);
 }
 
 static void _ocf_mngt_cache_init(ocf_cache_t cache,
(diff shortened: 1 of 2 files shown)


src/mngt/ocf_mngt_cache.c
```diff
@@ -1539,6 +1539,207 @@ struct ocf_pipeline_properties _ocf_mngt_cache_attach_pipeline_properties = {
 	},
 };
 
+typedef void (*_ocf_mngt_cache_unplug_end_t)(void *context, int error);
```
arutk: Please modify commit message to make it clear, that stop actually can return error code. However the only possible non-zero status is just a warning and cache is always stopped after executing ocf_mngt_cache_stop().

src/mngt/ocf_mngt_cache.c
```diff
@@ -2097,11 +2110,7 @@ void ocf_mngt_cache_stop(ocf_cache_t cache,
...
 
-	result = ocf_pipeline_create(&pipeline, cache,
-			&ocf_mngt_cache_stop_pipeline_properties);
-	if (result)
-		OCF_CMPL_RET(cache, priv, -OCF_ERR_NO_MEM);
-
+	pipeline = cache->stop_pipeline;
 	context = ocf_pipeline_get_priv(pipeline);
 
 	context->cmpl = cmpl;
```
arutk: Please replace error handling in line 2122 (env_strncpy) with BUG_ON - at this point we don't expect to receive any kind of error since all success conditions verified in advance. This would make it clear that stop does not return error code other than cache write error.

mmichal10:
retest this please

PR #339, opened 2020-01-15
Detect cache devices that would overflow ocf_cacheline_t

imjfckm:

### DIFF
--- a/src/metadata/metadata_hash.c
+++ b/src/metadata/metadata_hash.c
@@ -912,6 +912,7 @@ static int ocf_metadata_hash_init_variable_size(struct ocf_cache *cache,
 	struct ocf_cache_line_settings *settings =
 		(struct ocf_cache_line_settings *)&cache->metadata.settings;
 	ocf_flush_page_synch_t lock_page, unlock_page;
+	uint64_t device_lines;
 
 	OCF_DEBUG_TRACE(cache);
 
@@ -919,7 +920,16 @@ static int ocf_metadata_hash_init_variable_size(struct ocf_cache *cache,
 
 	ctrl = cache->metadata.iface_priv;
 
-	ctrl->device_lines = device_size / cache_line_size;
+	device_lines = device_size / cache_line_size;
+	if (device_lines >= (ocf_cache_line_t)(-1)){
+		/* TODO: This is just a rough check. Most optimal one would be
+		 * located in calculate_metadata_size. */
+		ocf_cache_log(cache, log_err, "Device exceeds maximum suported size "
+				"with this cache line size. Try bigger cache line size.");
+		return -OCF_ERR_INVAL_CACHE_DEV;
+	}
+
+	ctrl->device_lines = device_lines;
 
 	if (settings->size != cache_line_size)
 		/* Re-initialize settings with different cache line size */
--- a/tests/functional/tests/management/test_start_stop.py
+++ b/tests/functional/tests/management/test_start_stop.py
@@ -4,7 +4,7 @@
 #
 
 import logging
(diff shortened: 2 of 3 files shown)


tests/functional/tests/management/test_start_stop.py
```diff
@@ -330,6 +330,21 @@ def test_start_cache_same_id(pyocf_ctx, mode, cls):
     cache.get_stats()
 
 
+def test_start_cache_huge_device(pyocf_ctx):
```
Ostrokrzew: You didn't give the description for this test.
Also maybe You want to try parametrize it with a `cache line size`?

tests/functional/tests/management/test_start_stop.py
```diff
@@ -330,6 +330,20 @@ def test_start_cache_same_id(pyocf_ctx, mode, cls):
...
```
Ostrokrzew: Also maybe You want to try parametrize it with a `cache line size`?

tests/functional/tests/management/test_start_stop.py
imjfckm: Added

tests/functional/tests/management/test_start_stop.py
imjfckm: True that!

tests/functional/tests/management/test_start_stop.py
```diff
@@ -330,6 +330,31 @@ def test_start_cache_same_id(pyocf_ctx, mode, cls):
...
+@pytest.mark.parametrize("cls", CacheLineSize)
+def test_start_cache_huge_device(pyocf_ctx_log_buffer, cls):
+    """
+    Test whether we can start cache which would overflow ocf_cache_line_t type.
+    pass_criteria:
+      - Starting cache on device too big to handle should fail
+    """
+    class HugeDevice(Volume):
+        def get_length(self):
+            return Size.from_MiB(353888999)
```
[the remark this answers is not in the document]
robertbaldyga: It would be better to move this test to pyocf, where we can start cache with `metadata_volatile = true`. That way we can make it independent of metadata size.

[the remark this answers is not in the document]
imjfckm: Thanks! I think it's fixed now

src/metadata/metadata_hash.c
```diff
@@ -62,7 +62,7 @@ struct ocf_metadata_hash_ctrl {
 	ocf_cache_line_t cachelines;
 	ocf_cache_line_t start_page;
 	ocf_cache_line_t count_pages;
-	uint32_t device_lines;
+	uint64_t device_lines;
```
robertbaldyga: After changing type of `device_lines` to `uint64_t`, assignment in line 242 does value narrowing conversion to `int64_t`, which makes it possible for `cache_lines` to take negative values if cache device is huge enough.

imjfckm: Nice catch. I've changed `cache_lines` to `uint64_t` since it shouldn't ever go below zero.

robertbaldyga: So now it' the same about `diff_lines` and `lowest_diff`. Moveover in case of `diff_lines` we have possible multiplication overflow in line 298.

imjfckm: True that. I've reverted to previous version where we just check if we won't overflow on `device_lines`. This means there is a narrow window of device sizes which won't be supported even though they could. I think this close to the release I'm not brave enough to tune the `(...)calculate_metadata_size` function. Maybe next time? We should consider expanding ocf_cacheline_t anyways.

PR #342, opened 2020-01-23
Fix metadata flush

mmichal10:
Fixes https://github.com/Open-CAS/open-cas-linux/issues/253

### DIFF
--- a/src/metadata/metadata_status.h
+++ b/src/metadata/metadata_status.h
@@ -229,20 +229,23 @@ static inline bool metadata_clear_dirty_sec_changed(
 /*
  * Marks given cache line's bits as dirty
  *
- * @return true if the cache line was clean and became dirty
- * @return false if the cache line was dirty before marking bits
+ * @return true if any cache line's sector became dirty
+ * @return false for other cases
  */
 static inline bool metadata_set_dirty_sec_changed(
 		struct ocf_cache *cache, ocf_cache_line_t line,
-		uint8_t start, uint8_t stop)
+		uint8_t start, uint8_t stop, bool *line_was_dirty)
 {
-	bool was_dirty;
+	bool sec_changed;
 
 	OCF_METADATA_BITS_LOCK_WR();
-	was_dirty = cache->metadata.iface.set_dirty(cache, line, start, stop);
+	sec_changed = !cache->metadata.iface.test_dirty(cache, line,
+			start, stop, true);
+	*line_was_dirty = cache->metadata.iface.set_dirty(cache, line, start,
+			stop);
 	OCF_METADATA_BITS_UNLOCK_WR();
 
-	return !was_dirty;
+	return sec_changed;
 }
 
 /*******************************************************************************
--- a/src/metadata/metadata_status.h
+++ b/src/metadata/metadata_status.h
@@ -200,30 +200,25 @@ static inline bool metadata_test_and_clear_dirty_sec(
 /*
  * Marks given cache line's bits as clean
  *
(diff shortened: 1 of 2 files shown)


arutk:
Please add commit message explaining what exactly is being fixed and what was the failing scenario.

src/metadata/metadata_status.h
```diff
@@ -217,32 +218,37 @@ static inline bool metadata_clear_dirty_sec_changed(
...
+		*became_clean = !cache->metadata.iface.clear_dirty(cache, line,
 				start, stop);
 	}
 
 	OCF_METADATA_BITS_UNLOCK_WR();
 
-	return was_dirty && !is_dirty;
+	sec_changed = was_dirty;
+
+	return sec_changed;
```
robertbaldyga: Do we really need this additional variable?

mmichal10: fixed

src/metadata/metadata_status.h
```diff
@@ -200,14 +200,15 @@ static inline bool metadata_test_and_clear_dirty_sec(
...
  * Marks given cache line's bits as clean
  *
- * @return true if the cache line was dirty and became clean
+ * @return true if any cache line's sector was dirty and became clean
  * @return false for other cases
  */
 static inline bool metadata_clear_dirty_sec_changed(
 		struct ocf_cache *cache, ocf_cache_line_t line,
-		uint8_t start, uint8_t stop)
+		uint8_t start, uint8_t stop, bool *became_clean)
```
arutk: I suggest to revert the logic of last parameter and rename it to "line_was_clean", similarly to what metadata_set_dirty_sec_changed's  'line_was_dirty'.

This way the two functions are used similarily:
if (metadata_(clear/set)_dirty_sec_changed(..., &line_was(clear/dirty)) {
    mark_flush()
}
if (!line_was(clear/dirty)) {
   //udpate stats
}

src/metadata/metadata_status.h
```diff
@@ -217,32 +218,35 @@ static inline bool metadata_clear_dirty_sec_changed(
...
-	return was_dirty && !is_dirty;
+	return was_dirty;
 }
 
 /*
  * Marks given cache line's bits as dirty
  *
- * @return true if the cache line was clean and became dirty
- * @return false if the cache line was dirty before marking bits
+ * @return true if any cache line's sector became clean
```
michalwy: Wrong description, should be 'became dirty'.

src/metadata/metadata_status.h
mmichal10: Information about previous state of cache line is equivalent with information retrieved from metadata_clear_dirty_sec_changed() return value. Information that we need to have is new status of whole cache line  - if it became clean or not. And only if it did, we can change cache stats.

src/metadata/metadata_status.h
```diff
@@ -200,49 +200,55 @@ static inline bool metadata_test_and_clear_dirty_sec(
...
-			cache->metadata.settings.sector_start,
-			cache->metadata.settings.sector_end,
-			false);
+	sec_changed = cache->metadata.iface.test_dirty(cache, line,
+			start, stop, false);
 
-	if (was_dirty) {
-		is_dirty = cache->metadata.iface.clear_dirty(cache, line,
+	if (sec_changed) {
+		cache->metadata.iface.clear_dirty(cache, line,
```
mariuszbarczak: just take the result of it and pass negation to line_is_clean

PR #350, opened 2020-03-12
Add memory check before engaging promotion policy

imjfckm:

### DIFF
--- a/src/promotion/promotion.c
+++ b/src/promotion/promotion.c
@@ -48,6 +48,13 @@ ocf_error_t ocf_promotion_init(ocf_cache_t cache, ocf_promotion_t type)
 	if (result) {
 		env_vfree(cache->promotion_policy);
 		cache->promotion_policy = NULL;
+		ocf_cache_log(cache, log_info,
+				"Policy '%s' failed to initialize\n",
+				ocf_promotion_policies[type].name);
+	} else {
+		ocf_cache_log(cache, log_info,
+				"Policy '%s' initialized successfully\n",
+				ocf_promotion_policies[type].name);
 	}
 
 	return result;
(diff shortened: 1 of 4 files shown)


arutk:
When including this in OCL it would be nice to fix the env function to report available memory instead of free

src/promotion/promotion.c
```diff
@@ -83,6 +106,23 @@ ocf_error_t ocf_promotion_set_policy(ocf_promotion_policy_t policy,
...
+		if (size >= available) {
+			ocf_cache_log(cache, log_err, "Not enough memory to "
+					"initialize '%s' promotion policy! "
+					"Required %lu, available %lu. "
+					"Policy didn't change\n",
+					ocf_promotion_policies[type].name,
+					(long unsigned)size,
+					(long unsigned)available);
+
+			return -OCF_ERR_NO_MEM;
```
robertbaldyga: Is it enough for casadm to know exactly what message it should print? I guess `-OCF_ERR_NO_MEM` could happen also in other places (e.g. cache management lock can also return `-OCF_ERR_NO_MEM`). In case of cache start, when we don't have enough available memory we return `-OCF_ERR_NO_FREE_RAM` and it also seems to be better fit here.

src/promotion/ops.h
```diff
@@ -27,6 +27,9 @@ struct promotion_policy_ops {
 	void (*setup)(ocf_cache_t cache);
 		/*!< initialize promotion policy default config */
 
+	uint64_t (*size_of)(ocf_cache_t cache);
```
robertbaldyga: I have mixed feelings about making it a part of promotion policy ops. For now it seems to be problem of nhit only, so maybe it would be enough to print message from `init()` and then return appropriate error code.

src/promotion/promotion.c
imjfckm: True that, changed

src/promotion/ops.h
imjfckm: The extended API was meant to provide some fallback options, but it's changed since that so maybe it's not needed

PR #351, opened 2020-03-13
Fix seq_cutoff respecting in pt read

micrakow:
Fixes: https://github.com/Open-CAS/open-cas-linux/issues/345

### DIFF
--- a/src/engine/cache_engine.c
+++ b/src/engine/cache_engine.c
@@ -85,7 +85,7 @@ static const struct ocf_io_if IO_IFS[OCF_IO_PRIV_MAX_IF] = {
 		.write = ocf_discard,
 		.name = "Discard",
 	},
-	[OCF_IO_D2C_IF] =  {
+	[OCF_IO_D2C_IF] = {
 		.read = ocf_io_d2c,
 		.write = ocf_io_d2c,
 		.name = "Direct to core",
--- a/src/engine/cache_engine.c
+++ b/src/engine/cache_engine.c
@@ -245,6 +245,7 @@ void ocf_resolve_effective_cache_mode(ocf_cache_t cache,
 	if (ocf_seq_cutoff_check(core, req->rw, req->byte_position,
 				 req->byte_length)) {
 		req->cache_mode = ocf_req_cache_mode_pt;
+		req->seq_cutoff = 1;
 		return;
 	}
 
--- a/src/engine/cache_engine.c
+++ b/src/engine/cache_engine.c
@@ -223,36 +223,50 @@ void ocf_seq_cutoff_update(ocf_core_t core, struct ocf_request *req)
 	core->seq_cutoff.bytes += req->byte_length;
 }
 
-ocf_cache_mode_t ocf_get_effective_cache_mode(ocf_cache_t cache,
-		ocf_core_t core, struct ocf_io *io)
+void ocf_resolve_effective_cache_mode(ocf_cache_t cache,
+		ocf_core_t core, struct ocf_request *req)
 {
-	ocf_cache_mode_t mode;
+	if (req->d2c) {
+		req->cache_mode = ocf_req_cache_mode_d2c;
+		return;
+	}
 
-	if (cache->pt_unaligned_io && !ocf_req_is_4k(io->addr, io->bytes))
-		return ocf_cache_mode_pt;
+	if (ocf_fallback_pt_is_on(cache)){
+		req->cache_mode = ocf_req_cache_mode_pt;
+		return;
+	}
 
-	mode = ocf_part_get_cache_mode(cache,
-			ocf_part_class2id(cache, io->io_class));
-	if (!ocf_cache_mode_is_valid(mode))
-		mode = cache->conf_meta->cache_mode;
+	if (cache->pt_unaligned_io && !ocf_req_is_4k(req->byte_position,
+						     req->byte_length)) {
+		req->cache_mode = ocf_req_cache_mode_pt;
+		return;
+	}
 
-	if (ocf_seq_cutoff_check(core, io->dir, io->addr, io->bytes))
-		mode = ocf_cache_mode_pt;
+	if (ocf_seq_cutoff_check(core, req->rw, req->byte_position,
+				 req->byte_length)) {
+		req->cache_mode = ocf_req_cache_mode_pt;
+		return;
+	}
 
-	if (ocf_fallback_pt_is_on(cache))
-		mode = ocf_cache_mode_pt;
+	req->cache_mode = ocf_part_get_cache_mode(cache,
+				ocf_part_class2id(cache, req->part_id));
+	if (!ocf_cache_mode_is_valid(req->cache_mode))
+		req->cache_mode = cache->conf_meta->cache_mode;
 
-	return mode;
+	if (req->rw == OCF_WRITE &&
+	    ocf_req_cache_mode_has_lazy_write(req->cache_mode) &&
+	    ocf_req_set_dirty(req)) {
+		req->cache_mode = ocf_req_cache_mode_wt;
+	}
 }
 
-int ocf_engine_hndl_req(struct ocf_request *req,
-		ocf_req_cache_mode_t req_cache_mode)
+int ocf_engine_hndl_req(struct ocf_request *req)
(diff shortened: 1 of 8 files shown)


micrakow:
@CAS-Linux-Jenkins  retest this please

src/engine/cache_engine.c
```diff
@@ -223,36 +223,41 @@ void ocf_seq_cutoff_update(ocf_core_t core, struct ocf_request *req)
...
-	if (!ocf_cache_mode_is_valid(mode))
-		mode = cache->conf_meta->cache_mode;
+	if (cache->pt_unaligned_io && !ocf_req_is_4k(req->byte_position,
+						     req->byte_length)) {
+		req->cache_mode = ocf_req_cache_mode_pt;
+	}
 
-	if (ocf_seq_cutoff_check(core, io->dir, io->addr, io->bytes))
-		mode = ocf_cache_mode_pt;
+	mode = ocf_part_get_cache_mode(cache,
```
arutk: This information is lost. Did you mean req->cache_mode = ... ?

src/engine/cache_engine.c
```diff
@@ -223,36 +223,41 @@ void ocf_seq_cutoff_update(ocf_core_t core, struct ocf_request *req)
...
-		ocf_core_t core, struct ocf_io *io)
+void ocf_resolve_effective_cache_mode(ocf_cache_t cache,
+		ocf_core_t core, struct ocf_request *req)
 {
-	ocf_cache_mode_t mode;
+	ocf_cache_mode_t mode = ocf_cache_mode_none;
 
-	if (cache->pt_unaligned_io && !ocf_req_is_4k(io->addr, io->bytes))
-		return ocf_cache_mode_pt;
+	if (ocf_fallback_pt_is_on(cache)){
```
arutk: If you move this here, you should exit from the function after checking fallback PT - otherwise the cache mode might get overwritten by global cache mode settings.

src/engine/cache_engine.c
```diff
@@ -223,36 +223,52 @@ void ocf_seq_cutoff_update(ocf_core_t core, struct ocf_request *req)
...
+		req->seq_cutoff = 1;
+		return;
+	}
 
-	return mode;
+	req->cache_mode = ocf_part_get_cache_mode(cache,
+			ocf_part_class2id(cache, req->part_id));
+	if (!ocf_cache_mode_is_valid(req->cache_mode)) {
+		req->cache_mode = cache->conf_meta->cache_mode;
+		return;
```
arutk: to not return here

src/engine/cache_engine.c
```diff
@@ -223,36 +223,52 @@ void ocf_seq_cutoff_update(ocf_core_t core, struct ocf_request *req)
...
+	} else if (req->rw == OCF_WRITE &&
+	    ocf_req_cache_mode_has_lazy_write(req->cache_mode) &&
+	    ocf_io_set_dirty(req)) {
+		req->cache_mode = ocf_req_cache_mode_wt;
+		return;
+	}
```
arutk: Remove "else" - this should be checked regardless of previous condition

src/engine/cache_engine.c
```diff
@@ -223,36 +223,52 @@ void ocf_seq_cutoff_update(ocf_core_t core, struct ocf_request *req)
...
+	req->cache_mode = ocf_part_get_cache_mode(cache,
+			ocf_part_class2id(cache, req->part_id));
+	if (!ocf_cache_mode_is_valid(req->cache_mode)) {
+		req->cache_mode = cache->conf_meta->cache_mode;
+		return;
+	} else if (req->rw == OCF_WRITE &&
+	    ocf_req_cache_mode_has_lazy_write(req->cache_mode) &&
+	    ocf_io_set_dirty(req)) {
+		req->cache_mode = ocf_req_cache_mode_wt;
+		return;
```
arutk: I would remove this "return"

src/ocf_request.c
```diff
@@ -303,6 +303,12 @@ void ocf_req_put(struct ocf_request *req)
 	ocf_queue_put(queue);
 }
 
+int ocf_io_set_dirty(struct ocf_request *req)
```
arutk: "ocf_req_set_dirty" function name would better fit context and purpose

src/ocf_core.c
```diff
@@ -251,25 +243,14 @@ void ocf_core_submit_io_mode(struct ocf_io *io, ocf_cache_mode_t cache_mode)
 		return;
 	}
 
+	req->part_id = ocf_part_class2id(cache, io->io_class);
+	req->core = core;
+	req->complete = ocf_req_complete;
+
 	/* TODO: instead of casting ocf_cache_mode_t to ocf_req_cache_mode_t
 	   we can resolve IO interface here and get rid of the latter. */
```
arutk: Add:
req->cache_mode = cache_mode;

Pls check whether the same change applies to submit_fast.

src/ocf_core.c
```diff
@@ -251,25 +243,14 @@ void ocf_core_submit_io_mode(struct ocf_io *io, ocf_cache_mode_t cache_mode)
...
-	req_cache_mode = cache_mode;
-
 	if (cache_mode == ocf_cache_mode_none)
```
arutk: Move this check to ocf_resolve_effective_cache_mode  and call ocf_resolve_effective_cache_mode unconditionally.

Pls check whether the same change applies to submit_fast.

src/engine/cache_engine.c
```diff
@@ -223,36 +223,51 @@ void ocf_seq_cutoff_update(ocf_core_t core, struct ocf_request *req)
...
+		req->seq_cutoff = 1;
+		return;
+	}
 
-	if (ocf_fallback_pt_is_on(cache))
-		mode = ocf_cache_mode_pt;
+	req->cache_mode = ocf_part_get_cache_mode(cache,
+			ocf_part_class2id(cache, req->part_id));
+	if (!ocf_cache_mode_is_valid(req->cache_mode))
+		req->cache_mode = cache->conf_meta->cache_mode;
```
arutk: I think this is the block that should be executed conditionally
if (req->cache_mode == cache_mode_none)

arutk: I think this is the block that should be executed conditionally
if (req->cache_mode == cache_mode_none)

src/ocf_request.c
```diff
@@ -303,6 +303,12 @@ void ocf_req_put(struct ocf_request *req)
 	ocf_queue_put(queue);
 }
 
+int ocf_req_set_dirty(struct ocf_request *req)
+{
+	req->dirty = !!ocf_refcnt_inc(&req->cache->refcnt.dirty);
+	return req->dirty ? 0 : -EBUSY;
```
mmichal10: Shouldn't you use ocf internal error code?

src/engine/cache_engine.c
```diff
@@ -223,36 +223,51 @@ void ocf_seq_cutoff_update(ocf_core_t core, struct ocf_request *req)
...
-		mode = cache->conf_meta->cache_mode;
+	if (cache->pt_unaligned_io && !ocf_req_is_4k(req->byte_position,
+						     req->byte_length)) {
+		req->cache_mode = ocf_req_cache_mode_pt;
+		return;
+	}
 
-	if (ocf_seq_cutoff_check(core, io->dir, io->addr, io->bytes))
-		mode = ocf_cache_mode_pt;
+	if (ocf_seq_cutoff_check(core, req->rw, req->byte_position,
```
arutk: Seq cutoff will now stop fast path, which is not desired. The best solution would probably not to check for seq cutoff in submit_fast at all

micrakow: I'm dropping [resolving?] this issue according to our conversation.

src/ocf_core.c
```diff
@@ -223,10 +216,9 @@ static void ocf_req_complete(struct ocf_request *req, int error)
 	ocf_io_put(&req->ioi.io);
 }
 
-void ocf_core_submit_io_mode(struct ocf_io *io, ocf_cache_mode_t cache_mode)
```
robertbaldyga: Please remove it also from header.

PR #355, opened 2020-03-23
Fix stalls and warnings during flush

arutk:


src/mngt/ocf_mngt_flush.c
```diff
@@ -142,55 +142,61 @@ static int _ocf_mngt_get_sectors(ocf_cache_t cache, ocf_core_id_t core_id,
...
 
-		/* It's core_id cacheline and it's valid and it's dirty! */
-		p[j].cache_line = i;
-		p[j].core_line = core_line;
-		p[j].core_id = i_core_id;
-		j++;
-		/* stop if all cachelines were found */
-		if (j == dirty)
-			break;
+		if (++step % 100000 == 0) {
```
imjfckm: I know it's a very miniscule optimization, but its cost is miniscule as well - could you make it a power of two?

imjfckm: Also, is separate `step` variable really needed? It seems to be exact duplicate od `i`

src/mngt/ocf_mngt_flush.c
```diff
@@ -142,55 +142,61 @@ static int _ocf_mngt_get_sectors(ocf_cache_t cache, ocf_core_id_t core_id,
...
+		if (i_core_id == core_id &&
+				metadata_test_valid_any(cache, i) &&
+				metadata_test_dirty(cache, i)) {
+			/* It's valid and dirty target core cacheline */
+			p[j].cache_line = i;
+			p[j].core_line = core_line;
+			p[j].core_id = i_core_id;
+			j++;
+
+			/* stop if all cachelines were found */
```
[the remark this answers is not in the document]
arutk: Renaming souds good to me. But I would leave the comment anyway.

src/mngt/ocf_mngt_flush.c
arutk: All done

src/mngt/ocf_mngt_flush.c
```diff
@@ -141,56 +141,61 @@ static int _ocf_mngt_get_sectors(ocf_cache_t cache, ocf_core_id_t core_id,
 	uint64_t core_line;
 	ocf_core_id_t i_core_id;
 	struct flush_data *p;
-	uint32_t i, j, dirty = 0;
+	uint32_t i, dirty_found = 0, dirty_total = 0;
```
mmichal10: `i` is referring do cache_line_id? Maybe it would be also worth renaming? The same for `p` - if you have any better ideas to name it.

src/mngt/ocf_mngt_flush.c
```diff
@@ -141,56 +141,61 @@ static int _ocf_mngt_get_sectors(ocf_cache_t cache, ocf_core_id_t core_id,
...
-			continue;
-
-		if (!metadata_test_dirty(cache, i))
-			continue;
-
-		if (ocf_cache_line_is_used(cache, i))
-			continue;
+		if (i_core_id == core_id &&
+				metadata_test_valid_any(cache, i) &&
+				metadata_test_dirty(cache, i)) {
```
mmichal10: How about warping it within macro or inline function `cache_line_is_dirty`?

arutk: I would rather not, now these functions have a single purpose and the names are descriptive. cache_line_is_dirty would easily be confused with cacheline_test_dirty .

PR #359, opened 2020-03-27
Reject discard IOs with size = 0

### LINKED ISSUE #353: OCF breaks on discard with size=0
This problem occurred on OCF v19.3. Not checked on current master.

Ostrokrzew:
Reject zero-size IOs - return *invalid input parameter value* error.
Add test for above case.
Fix some typos.
Remove unused parameters from code.

Fix #353

### DIFF
--- a/src/engine/cache_engine.h
+++ b/src/engine/cache_engine.h
@@ -66,8 +66,7 @@ static inline bool ocf_req_cache_mode_has_lazy_write(ocf_req_cache_mode_t mode)
 
 bool ocf_fallback_pt_is_on(ocf_cache_t cache);
 
-struct ocf_request *ocf_engine_pop_req(struct ocf_cache *cache,
-		struct ocf_queue *q);
+struct ocf_request *ocf_engine_pop_req(struct ocf_queue *q);
 
 int ocf_engine_hndl_req(struct ocf_request *req);
 
--- a/src/ocf_core.c
+++ b/src/ocf_core.c
@@ -392,6 +392,11 @@ static void ocf_core_volume_submit_discard(struct ocf_io *io)
 
 	OCF_CHECK_NULL(io);
 
+	if (io->bytes == 0) {
+		ocf_io_end(io, -OCF_ERR_INVAL);
+		return;
+	}
+
 	ret = ocf_core_validate_io(io);
 	if (ret < 0) {
 		ocf_io_end(io, ret);
--- a/inc/ocf_metadata.h
+++ b/inc/ocf_metadata.h
@@ -28,7 +28,7 @@ struct ocf_atomic_metadata {
 	/** Set bit indicates that given sector is valid (is cached) */
 	uint32_t valid : 1;
 
-	/** Set bit indicates that sector i dirty */
+	/** Set bit indicates that sector is dirty */
 	uint32_t dirty : 1;
 } __attribute__((packed));
 
--- a/src/engine/cache_engine.c
+++ b/src/engine/cache_engine.c
@@ -116,7 +116,7 @@ const struct ocf_io_if *ocf_get_io_if(ocf_req_cache_mode_t req_cache_mode)
 	return cache_mode_io_if_map[req_cache_mode];
 }
 
-struct ocf_request *ocf_engine_pop_req(ocf_cache_t cache, ocf_queue_t q)
+struct ocf_request *ocf_engine_pop_req(ocf_queue_t q)
 {
 	unsigned long lock_flags = 0;
 	struct ocf_request *req;
--- a/src/ocf_queue.c
+++ b/src/ocf_queue.c
@@ -86,13 +86,10 @@ void ocf_io_handle(struct ocf_io *io, void *opaque)
 void ocf_queue_run_single(ocf_queue_t q)
 {
 	struct ocf_request *io_req = NULL;
-	ocf_cache_t cache;
 
 	OCF_CHECK_NULL(q);
 
-	cache = q->cache;
-
-	io_req = ocf_engine_pop_req(cache, q);
+	io_req = ocf_engine_pop_req(q);
 
 	if (!io_req)
 		return;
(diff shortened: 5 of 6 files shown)


inc/ocf_io.h
```diff
@@ -226,7 +226,7 @@ ctx_data_t *ocf_io_get_data(struct ocf_io *io);
  * @param[in] io OCF IO to be handled
  * @param[in] opaque OCF opaque
  */
-void ocf_io_handle(struct ocf_io *io, void *opaque);
+void ocf_io_handle(void *opaque);
```
imjfckm: If anything I think that `io` is the param that should stay ;) Afterall - it's `io_handle`

tests/functional/tests/engine/test_io_zero_size.py
```diff
@@ -0,0 +1,45 @@
...
+
+    cmpl = OcfCompletion([("err", c_int)])
+    io.callback = cmpl.callback
+    io.submit()
+    cmpl.wait()
+
+    assert cmpl.results["err"] != 0
+    assert cache_device.get_stats()[IoDir.WRITE] == 0
+    stats = cache.get_stats()
+    assert stats["req"]["wr_full_misses"]["value"] == 0
```
imjfckm: I don't think that misses should be checked here - it's only obscuring the more important asserts

tests/functional/tests/engine/test_io_zero_size.py
```diff
@@ -0,0 +1,45 @@
...
+    assert stats["usage"]["occupancy"]["value"] == 0
+
+    assert core.exp_obj_md5() == core_device.md5()
```
imjfckm: I'm sure ths check here is redundand as well - we don't need to check the data consistency if no IO was issued

tests/functional/tests/engine/test_io_zero_size.py
```diff
@@ -0,0 +1,45 @@
...
+    io = core.new_core_io(cache.get_default_queue(), Size.from_sector(1).B,
+                          0, IoDir.WRITE, 0, 0)
+    io.set_data(0)
+
+    cmpl = OcfCompletion([("err", c_int)])
+    io.callback = cmpl.callback
+    io.submit()
+    cmpl.wait()
+
+    assert cmpl.results["err"] != 0
```
robertbaldyga: We could even check for specific error code.

tests/functional/tests/engine/test_io_zero_size.py
```diff
@@ -0,0 +1,45 @@
...
+                          0, IoDir.WRITE, 0, 0)
+    io.set_data(0)
+
+    cmpl = OcfCompletion([("err", c_int)])
+    io.callback = cmpl.callback
+
+    io.submit()
+    cmpl.wait()
+
+    assert cmpl.results["err"] == -5  # I/O error
```
robertbaldyga: Does this test pass? Because in this scenario CAS should return `-OCF_ERR_INVAL` which has numerical value of -1000000, not -5.

Ostrokrzew: Should, but this code's error has value of -5, which is I/O error in `errno-base.h`. See here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/asm-generic/errno-base.h
I don't know why, because it's not returned anywhere in OCF's code. I'm checking it.

robertbaldyga: `OCF_ERR_INVAL` is defined in [https://github.com/Open-CAS/ocf/blob/master/inc/ocf_err.h](https://github.com/Open-CAS/ocf/blob/master/inc/ocf_err.h) and definitely it doesn't have value -5.

Ostrokrzew: The test was written badly, now it works and the error code is correct.

src/ocf_core.c
```diff
@@ -181,6 +181,9 @@ static inline int ocf_core_validate_io(struct ocf_io *io)
 	if (io->addr + io->bytes > ocf_volume_get_length(volume))
 		return -OCF_ERR_INVAL;
 
+	if (io->bytes == 0)
+		return -OCF_ERR_INVAL;
+
```
arutk: I think it's expected to receive e.g. write request with size == 0 - it would make sense for example if the I/O has FLUSH flag set. Let's analyze this more carefully before we introduce this change.

src/ocf_core.c
```diff
@@ -173,14 +173,17 @@ static inline void dec_counter_if_req_was_dirty(struct ocf_request *req)
...
 {
 	ocf_volume_t volume = ocf_io_get_volume(io);
 	ocf_core_t core = ocf_volume_to_core(volume);
 
 	if (io->addr + io->bytes > ocf_volume_get_length(volume))
 		return -OCF_ERR_INVAL;
 
+	if (io->bytes == 0 && discard)
+		return -OCF_ERR_INVAL;
+
```
robertbaldyga, line 186: Why not just move this to `ocf_core_volume_submit_discard()`?

src/engine/cache_engine.h
```diff
@@ -66,8 +66,10 @@ static inline bool ocf_req_cache_mode_has_lazy_write(ocf_req_cache_mode_t mode)
 
 bool ocf_fallback_pt_is_on(ocf_cache_t cache);
 
-struct ocf_request *ocf_engine_pop_req(struct ocf_cache *cache,
-		struct ocf_queue *q);
+bool ocf_seq_cutoff_check(ocf_core_t core, uint32_t dir, uint64_t addr,
+		uint64_t bytes);
```
robertbaldyga, line 70: There is no such function in OCF. Please remove it from the header.

Ostrokrzew: Sorry, I don't know how it got here.

PR #364, opened 2020-04-15
Force pass-through for requests bigger than cache

### LINKED ISSUE #200: Eviction occurs when IO greater than cache size is submitted.
Eviction is triggered upon writing to cache with request bigger than cache size.
IO should go directly to core and eviction should not occur. 
**Test for this particular case can be found in #199**

micrakow:
Fixes #200

### DIFF
--- a/src/engine/cache_engine.c
+++ b/src/engine/cache_engine.c
@@ -181,6 +181,11 @@ void ocf_resolve_effective_cache_mode(ocf_cache_t cache,
 		return;
 	}
 
+	if (req->core_line_count > cache->conf_meta->cachelines) {
+		req->cache_mode = ocf_req_cache_mode_pt;
+		return;
+	}
+
 	if (ocf_core_seq_cutoff_check(core, req)) {
 		req->cache_mode = ocf_req_cache_mode_pt;
 		req->seq_cutoff = 1;
--- a/src/ocf_seq_cutoff.c
+++ b/src/ocf_seq_cutoff.c
@@ -7,17 +7,19 @@
 #include "ocf_cache_priv.h"
 #include "ocf_priv.h"
 #include "ocf/ocf_debug.h"
+#include "utils/utils_cache_line.h"
 
 #define SEQ_CUTOFF_FULL_MARGIN \
                 (OCF_TO_EVICTION_MIN + OCF_PENDING_EVICTION_LIMIT)
 
-static inline bool ocf_seq_cutoff_is_on(ocf_cache_t cache)
+static inline bool ocf_seq_cutoff_is_on(ocf_cache_t cache,
+		struct ocf_request *req)
 {
 	if (!ocf_cache_is_device_attached(cache))
 		return false;
 
 	return (ocf_freelist_num_free(cache->freelist) <=
-			SEQ_CUTOFF_FULL_MARGIN);
+				SEQ_CUTOFF_FULL_MARGIN + req->core_line_count);
 }
 
 static int ocf_seq_cutoff_stream_cmp(struct ocf_rb_node *n1,
@@ -101,9 +103,8 @@ bool ocf_core_seq_cutoff_check(ocf_core_t core, struct ocf_request *req)
 	switch (policy) {
 		case ocf_seq_cutoff_policy_always:
 			break;
-
 		case ocf_seq_cutoff_policy_full:
-			if (ocf_seq_cutoff_is_on(cache))
+			if (ocf_seq_cutoff_is_on(cache, req))
 				break;
 			return false;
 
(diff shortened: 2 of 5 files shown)


tests/functional/tests/eviction/test_eviction.py
```diff
@@ -21,8 +21,7 @@
 
 @pytest.mark.parametrize("cls", CacheLineSize)
 @pytest.mark.parametrize("mode", [CacheMode.WT, CacheMode.WB, CacheMode.WO])
-@pytest.mark.xfail  # TODO: remove when fixed
-def test_write_size_greater_than_cache(pyocf_ctx, mode: CacheMode, cls: CacheLineSize):
+def test_orig(pyocf_ctx, mode: CacheMode, cls: CacheLineSize):
```
arutk, line 24: pls fix test name

tests/functional/tests/eviction/test_eviction.py
```diff
@@ -21,8 +21,7 @@
...
     """Test if eviction does not occur when IO greater than cache size is submitted.
     """
     cache_device = Volume(Size.from_MiB(20))  # this gives about 1.375 MiB actual caching space
```
arutk, line 27: I suggest to modify the test to collect cache capacity from the stats rather than assuming the value of 1.375

tests/functional/tests/eviction/test_eviction.py
```diff
@@ -39,35 +38,40 @@ def test_write_size_greater_than_cache(pyocf_ctx, mode: CacheMode, cls: CacheLin
...
-        stats["block"]["cache_volume_rd"]["value"] + \
-        prev_writes_to_core + io_size_bigger_than_cache.B / Size.from_KiB(4).B, \
-        "Writes to core after second IO"
-
-    # Occupancy shouldn't change (no eviction)
-    assert stats["usage"]["occupancy"]["value"] == (valid_io_size.B / Size.from_KiB(4).B),\
-        "Occupancy after second IO"
+    # Second write shouldn't affect cache and should go directly to core.
+    # Cache occupancy shouldn't change
+    assert first_usage_sts['occupancy'] == \
```
arutk, line 64: I would also check stats to make sure the last I/O was sent in PT

src/engine/cache_engine.c
```diff
@@ -242,6 +251,11 @@ void ocf_resolve_effective_cache_mode(ocf_cache_t cache,
 		return;
 	}
 
+	if(req->core_line_count > cache->conf_meta->cachelines){
```
arutk, line 254: missing whitespaces

src/engine/cache_engine.c
```diff
@@ -164,12 +165,20 @@ bool ocf_fallback_pt_is_on(ocf_cache_t cache)
...
+ *
+ * @return true if seq cutoff should be activated, false otherwise
+ */
+static inline bool ocf_seq_cutoff_is_on(ocf_cache_t cache, uint64_t clines)
 {
 	if (!ocf_cache_is_device_attached(cache))
 		return false;
 
-	return (ocf_freelist_num_free(cache->freelist) <=
+	return ((ocf_freelist_num_free(cache->freelist) - clines) <=
```
arutk, line 181: This may cause signed integer underflow - would be better to move "-clines" to the right side of "<="

src/ocf_seq_cutoff.c
```diff
@@ -7,17 +7,26 @@
...
+ * @return true if seq cutoff should be activated, false otherwise
+ */
+static inline bool ocf_seq_cutoff_is_on(ocf_cache_t cache, uint64_t clines)
 {
 	if (!ocf_cache_is_device_attached(cache))
 		return false;
 
 	return (ocf_freelist_num_free(cache->freelist) <=
-			SEQ_CUTOFF_FULL_MARGIN);
+				(SEQ_CUTOFF_FULL_MARGIN - clines));
```
arutk, line 29: 1. integer underflow is very likely here
2. 'clines' sign seems to be inverted - should be "... <= (SEQ_CUTOFF_FULL_MARGIN + clines)
3. i think brackets around "(SEQ_CUTOFF_FULL_MARGIN - clines)" are unnecessary

src/ocf_seq_cutoff.c
```diff
@@ -7,17 +7,26 @@
...
-static inline bool ocf_seq_cutoff_is_on(ocf_cache_t cache)
+/**
+ * @brief Check if sequential cutoff should be activated or not
+ *
+ * @param cache Cache instance
+ * @param clines Cachelines number that could be inserted with current request
+ *
+ * @return true if seq cutoff should be activated, false otherwise
+ */
+static inline bool ocf_seq_cutoff_is_on(ocf_cache_t cache, uint64_t clines)
```
robertbaldyga, line 23: This function could take only one argument `req` and get number of clines from `req->core_line_count`. That would simplify this function call in `ocf_core_seq_cutoff_check()`.

PR #371, opened 2020-05-20
Disable loading cache with 'force' flag

### LINKED ISSUE #361: Loading cache with force set to true fails if cache was shutdown without flush
### Description
When trying to load dirty shutdown wb cache it is possible to use `force` flag, which bypasses the `dirty_shutdown` bit check in attach and then fails the CRC check for `collision` segment.

### Suggested solution
Reject attempt to load the cache with `force` flag set, or silently ignore it - it doesn't make sense in this context.

Ostrokrzew:
Fail `ocf_mngt_cache_load` function with `EINVAL` error code
when force flag is in use.
Log error message.

Closes #361

Signed-off-by: Slawomir Jankowski <slawomir.jankowski@intel.com>

### DIFF
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -2075,7 +2075,7 @@ void ocf_mngt_cache_load(ocf_cache_t cache,
 
 	/* Load is not allowed in volatile metadata mode */
 	if (cache->metadata.is_volatile)
-		OCF_CMPL_RET(cache, priv, -EINVAL);
+		OCF_CMPL_RET(cache, priv, -OCF_ERR_INVAL);
 
 	/* Load is not allowed with 'force' flag on */
 	if (cfg->force) {
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -2077,6 +2077,13 @@ void ocf_mngt_cache_load(ocf_cache_t cache,
 	if (cache->metadata.is_volatile)
 		OCF_CMPL_RET(cache, priv, -EINVAL);
 
+	/* Load is not allowed with 'force' flag on */
+	if (cfg->force) {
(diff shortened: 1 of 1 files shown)


src/mngt/ocf_mngt_cache.c
```diff
@@ -2077,6 +2077,13 @@ void ocf_mngt_cache_load(ocf_cache_t cache,
 	if (cache->metadata.is_volatile)
 		OCF_CMPL_RET(cache, priv, -EINVAL);
 
+	/* Load is not allowed with 'force' flag on */
+	if (cfg->force) {
+		ocf_cache_log(cache, log_err, "Starting cache with 'load' "
+				"and 'force' options simultaneously is forbidden.");
+		OCF_CMPL_RET(cache, priv, -EINVAL);
```
imjfckm, line 2084: `EINVAL` is not a valid OCF return code - but I kinda see why you put it here :) could you fix both?

Ostrokrzew: point for You

src/mngt/ocf_mngt_cache.c
```diff
@@ -2075,7 +2075,14 @@ void ocf_mngt_cache_load(ocf_cache_t cache,
 
 	/* Load is not allowed in volatile metadata mode */
 	if (cache->metadata.is_volatile)
-		OCF_CMPL_RET(cache, priv, -EINVAL);
+		OCF_CMPL_RET(cache, priv, -OCF_ERR_INVAL);
+
+	/* Load is not allowed with 'force' flag on */
+	if (cfg->force) {
+		ocf_cache_log(cache, log_err, "Starting cache with 'load' "
+				"and 'force' options simultaneously is forbidden.");
```
robertbaldyga, line 2083: Actually from OCF point of view "load" is not an option - is operation that has it's own dedicated function. So I'd reword this message to something like "Using 'force' flag is forbidden for load operation".

PR #376, opened 2020-06-08
Per-partition max occupancy

micrakow:
Respect max occupancy for each partition.
Needed cachelines number is evicted after max occupancy threshold is reached.

### DIFF
--- a/inc/ocf_io_class.h
+++ b/inc/ocf_io_class.h
@@ -40,8 +40,8 @@ struct ocf_io_class_info {
 
 	uint32_t max_size;
 		/*!< Maximum number of cache lines that might be assigned into
-		 * this IO class. If current size reach maximum size no more
-		 * allocation for this IO class takes place
+		 * this IO class. If current size reach maximum size then some
+		 * of ioclass's cachelines are evicted.
 		 */
 
 	uint8_t eviction_policy_type;
--- a/src/engine/engine_common.c
+++ b/src/engine/engine_common.c
@@ -14,6 +14,7 @@
 #include "../utils/utils_cache_line.h"
 #include "../ocf_request.h"
 #include "../utils/utils_cleaner.h"
+#include "../utils/utils_part.h"
 #include "../metadata/metadata.h"
 #include "../eviction/eviction.h"
 #include "../promotion/promotion.h"
@@ -464,6 +465,20 @@ int ocf_engine_prepare_clines(struct ocf_request *req,
 	if (mapped || !promote) {
 		/* Will not attempt mapping - release hash bucket lock */
 		ocf_req_hash_unlock_rd(req);
+	} else if (ocf_part_occ_is_limit_reached(req)) {
+		/* Partition reaches its max occupnacy, some cachelines
+		 * must be evcited first */
+		req->part_evict = true;
+		ocf_req_hash_unlock_rd(req);
+		ocf_metadata_start_exclusive_access(metadata_lock);
+
+		if (ocf_engine_evict(req) == LOOKUP_MAPPED)
+			ocf_engine_map(req);
+
+		if (!req->info.mapping_error)
+			lock = lock_clines(req, engine_cbs);
+
+		ocf_metadata_end_exclusive_access(metadata_lock);
 	} else {
 		/* Need to map (potentially evict) cachelines. Mapping must be
 		 * performed holding (at least) hash-bucket write lock */
--- a/src/eviction/eviction.c
+++ b/src/eviction/eviction.c
@@ -42,12 +42,17 @@ static uint32_t ocf_evict_calculate(struct ocf_user_part *part,
 
 static inline uint32_t ocf_evict_do(ocf_cache_t cache,
 		ocf_queue_t io_queue, const uint32_t evict_cline_no,
-		ocf_part_id_t target_part_id)
+		ocf_part_id_t target_part_id, bool part_evict)
 {
 	uint32_t to_evict = 0, evicted = 0;
 	struct ocf_user_part *part;
 	struct ocf_user_part *target_part = &cache->user_parts[target_part_id];
 	ocf_part_id_t part_id;
+
+	if (part_evict) {
+		return ocf_eviction_need_space(cache, io_queue,
+				target_part_id, evict_cline_no);
+	}
 
 	/* For each partition from the lowest priority to highest one */
 	for_each_part(cache, part, part_id) {
@@ -110,12 +115,13 @@ int space_managment_evict_do(struct ocf_cache *cache,
 	uint32_t free;
 
 	free = ocf_freelist_num_free(cache->freelist);
-	if (evict_cline_no <= free)
+	if (!req->part_evict && evict_cline_no <= free)
 		return LOOKUP_MAPPED;
+	else if (!req->part_evict)
+		evict_cline_no -= free;
 
-	evict_cline_no -= free;
 	evicted = ocf_evict_do(cache, req->io_queue, evict_cline_no,
-			req->part_id);
+			req->part_id, req->part_evict);
 
 	if (evict_cline_no <= evicted)
 		return LOOKUP_MAPPED;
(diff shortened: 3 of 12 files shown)


src/utils/utils_part.h
```diff
@@ -61,7 +62,26 @@ static inline void ocf_part_sort(struct ocf_cache *cache)
 	ocf_lst_sort(&cache->lst_part);
 }
 
-static inline ocf_cache_mode_t ocf_part_get_cache_mode(struct ocf_cache *cache,
+static inline bool ocf_part_occ_check_limit(struct ocf_request *req)
```
Ostrokrzew, line 65: This name could be confusing (for me it is), maybe rephrase it to something like `is_ocf_part_occ_reach_limit`.

src/utils/utils_part.h
```diff
@@ -61,7 +62,26 @@ static inline void ocf_part_sort(struct ocf_cache *cache)
...
+		/* Partition reaches its max occupnacy, some cachelines
+		 * must be evcited first */
+		req->part_evict = true;
+		return true;
+	} else
+		return false;
+}
+
+
+static inline ocf_cache_mode_t ocf_part_get_cache_mode(ocf_cache_t cache,
```
Ostrokrzew, line 84: Did You change parameter on purpose?

inc/ocf_io_class.h
```diff
@@ -40,7 +40,8 @@ struct ocf_io_class_info {
 
 	uint32_t max_size;
 		/*!< Maximum number of cache lines that might be assigned into
-		 * this IO class. If current size reach maximum size no more
+		 * this IO class. If current size reach maximum size then some
+		 * of ioclass's cachelines are evicted.
 		 * allocation for this IO class takes place
```
Ostrokrzew, line 45: You probably left this line by mistake. It's inconsistent with what You wrote.

src/ocf_request.h
```diff
@@ -188,6 +188,9 @@ struct ocf_request {
 	uint8_t seq_cutoff : 1;
 	/*!< Sequential cut off set for this request */
 
+	/* !< Some cachelines from request's partition must be evicted */
+	uint8_t part_evict : 1;
```
Ostrokrzew, line 192: To remain consistency, this comment should be under variable.

src/utils/utils_part.h
micrakow: Yes I did! :)

KamilLepek:
retest this please

KamilLepek:
retest this please

src/utils/utils_part.h
```diff
@@ -61,7 +62,26 @@ static inline void ocf_part_sort(struct ocf_cache *cache)
...
+	uint32_t clines = ocf_engine_unmapped_count(req);
+	uint32_t max = req->cache->user_parts[req->part_id].config->max_size;
+	uint32_t curr_cached = env_atomic_read(
+				&cache->user_parts[req->part_id].runtime->valid_cnt);
+
+	uint32_t test = curr_cached + clines;
+	if (max && test > max) {
+		/* Partition reaches its max occupnacy, some cachelines
+		 * must be evcited first */
+		req->part_evict = true;
```
mmichal10, line 77: Function name doesn't indicate it has any side effects like setting flag. I believe moving to caller function would make code more intuitive

src/eviction/eviction.c
```diff
@@ -42,60 +42,66 @@ static uint32_t ocf_evict_calculate(struct ocf_user_part *part,
...
 	uint32_t to_evict = 0, evicted = 0;
 	struct ocf_user_part *part;
 	struct ocf_user_part *target_part = &cache->user_parts[target_part_id];
 	ocf_part_id_t part_id;
 
-	/* For each partition from the lowest priority to highest one */
-	for_each_part(cache, part, part_id) {
+	if (part_evict) {
+		evicted = ocf_eviction_need_space(cache, io_queue,
+				target_part_id, evict_cline_no);
```
mmichal10, line 54: Returning value here will let You avoid using `else` and making another indentation block

src/engine/engine_common.c
```diff
@@ -465,29 +466,42 @@ int ocf_engine_prepare_clines(struct ocf_request *req,
...
-		ocf_req_hash_lock_upgrade(req);
-		ocf_engine_map(req);
-		if (!req->info.mapping_error)
-			lock = lock_clines(req, engine_cbs);
-		ocf_req_hash_unlock_wr(req);
-
-		if (req->info.mapping_error) {
-			/* Not mapped - evict cachelines under global exclusive
-			 * lock*/
+		if (ocf_part_occ_check_limit(req)) {
```
robertbaldyga, line 469: That might be `else if` instead of `if` nested within `else` block.

src/utils/utils_part.h
robertbaldyga, line 77: Setting `req->part_evict` in this function is probably not what most of the people would expect. Please move it to engine_common.c to make code flow more obvious.

src/utils/utils_part.h
```diff
@@ -61,7 +62,26 @@ static inline void ocf_part_sort(struct ocf_cache *cache)
...
+		return true;
+	} else
```
robertbaldyga, line 79: No need for `else` when `if` always returns.

src/utils/utils_part.h
```diff
@@ -61,7 +62,26 @@ static inline void ocf_part_sort(struct ocf_cache *cache)
...
-static inline ocf_cache_mode_t ocf_part_get_cache_mode(struct ocf_cache *cache,
+static inline bool ocf_part_occ_check_limit(struct ocf_request *req)
+{
+	ocf_cache_t cache = req->cache;
+	uint32_t clines = ocf_engine_unmapped_count(req);
+	uint32_t max = req->cache->user_parts[req->part_id].config->max_size;
+	uint32_t curr_cached = env_atomic_read(
+				&cache->user_parts[req->part_id].runtime->valid_cnt);
+
+	uint32_t test = curr_cached + clines;
```
robertbaldyga, line 73: This is hard to understand why this variable is named `test`. I'd make this entire check like this:
```c
uint32_t available_space = max ? max - curr_cached : 0;

if (clines > available_space) {
	...
```
BTW empty line between declarations and code is required.

src/engine/engine_common.c
```diff
@@ -464,31 +465,45 @@ int ocf_engine_prepare_clines(struct ocf_request *req,
...
+				 * traverse once again and evict cachelines if needed.
+				 */
+				if (ocf_engine_evict(req) == LOOKUP_MAPPED)
+					ocf_engine_map(req);
+
+				if (!req->info.mapping_error)
+					lock = lock_clines(req, engine_cbs);
+
+				ocf_metadata_end_exclusive_access(metadata_lock);
+			}
```
robertbaldyga, line 506: This code block is overindented.

src/eviction/eviction.c
```diff
@@ -42,13 +42,17 @@ static uint32_t ocf_evict_calculate(struct ocf_user_part *part,
...
+		ocf_part_id_t target_part_id, bool part_evict)
 {
 	uint32_t to_evict = 0, evicted = 0;
 	struct ocf_user_part *part;
 	struct ocf_user_part *target_part = &cache->user_parts[target_part_id];
 	ocf_part_id_t part_id;
 
+	if (part_evict)
+		return ocf_eviction_need_space(cache, io_queue,
+				target_part_id, evict_cline_no);
```
robertbaldyga, line 54: For non-one-line code block please use brackets.

src/utils/utils_part.h
robertbaldyga: Yep, when I look at it I'm not sure if it should return `true` or `false` when the limit is reached. The name should indicate what the result means, not only what is checked.

src/ocf_io_class.c
```diff
@@ -33,9 +34,11 @@ int ocf_cache_io_class_get_info(ocf_cache_t cache, uint32_t io_class,
 		return -OCF_ERR_INVAL;
 	}
 
+	if (ocf_cache_is_device_attached(cache))
+		cache_occupancy_total = env_atomic_read(&cache->user_parts[part_id].runtime->valid_cnt);
```
robertbaldyga, line 38: This line is way over 80 chars.

PR #385, opened 2020-07-16
Two pass write invalidate

arutk:

### DIFF
--- a/src/engine/engine_wi.c
+++ b/src/engine/engine_wi.c
@@ -18,9 +18,41 @@
 
 static int ocf_write_wi_update_and_flush_metadata(struct ocf_request *req);
 
-static const struct ocf_io_if _io_if_wi_flush_metadata = {
+static const struct ocf_io_if _io_if_wi_update_metadata = {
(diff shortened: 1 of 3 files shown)


mmichal10:
retest this please

arutk:
retest this please

mmichal10:
retest this please

src/engine/engine_wi.c
```diff
@@ -18,11 +18,49 @@
...
+			mark it as completed */
+		req->wi_second_pass = true;
+	}
+
+	if (req->wi_second_pass) {
+		req->complete(req, req->error);
+		ocf_req_put(req);
+
+		return 0;
+	}
```
robertbaldyga, line 41: This is not needed, as we already checked for miss before entering this function.

arutk: This check is needed when this function called from IO queue after metadata flush

arutk: hit check is indeed seems to be redundant

PR #394, opened 2020-07-23
Make CRC mismatch on recovery a warning instead of error

robertbaldyga:

### DIFF
--- a/src/metadata/metadata_hash.c
+++ b/src/metadata/metadata_hash.c
@@ -1274,7 +1274,9 @@ static void ocf_medatata_hash_check_crc_sb_config(ocf_pipeline_t pipeline,
 }
 
 static void ocf_medatata_hash_check_crc_skip(ocf_pipeline_t pipeline,
-		void *priv, ocf_pipeline_arg_t arg, bool skip_on_dirty_shutdown)
+		void *priv, ocf_pipeline_arg_t arg,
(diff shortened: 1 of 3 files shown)


src/metadata/metadata_hash.c
```diff
@@ -1293,10 +1295,16 @@ static void ocf_medatata_hash_check_crc_skip(ocf_pipeline_t pipeline,
 
 	if (crc != sb_config->checksum[segment]) {
 		/* Checksum does not match */
-		ocf_cache_log(cache, log_err,
-				"Loading %s ERROR, invalid checksum",
-				ocf_metadata_hash_raw_names[segment]);
-		OCF_PL_FINISH_RET(pipeline, -OCF_ERR_INVAL);
+		if (warning_only) {
```
arutk, line 1298: warning_only && !sb_config->clean_shutdown ?

robertbaldyga: This doesn't need to be in function. It depends on the context and we want to be it context bound instead of having it hardcoded in function which only checks CRC. The only use of warning_only variant is in recovery path anyway.

arutk: in current implementation warning_only is not limited to recovery path.
ocf_medatata_hash_check_crc_warning calls this function with warning_only regardless of shutdown status.

PR #399, opened 2020-08-19
Fix memory requirement when loading cache

rafalste:
- supply _ocf_mngt_calculate_ram_needed() function with just the values it actually uses instead of whole structs
- load properties before checking memory needs and obtain cache line size from context rather than from cache state

Fixes https://github.com/Open-CAS/open-cas-linux/issues/413


imjfckm:
@CAS-Linux-Jenkins add to whitelist

Ostrokrzew:
add to whitelist

Ostrokrzew (approved):

imjfckm (approved):

PR #413, opened 2020-11-06
Occupancy per ioclass

mmichal10:


src/metadata/metadata_hash.c
```diff
@@ -563,6 +563,7 @@ int ocf_metadata_hash_init(struct ocf_cache *cache,
 	for (i = 0; i < OCF_IO_CLASS_MAX + 1; i++) {
 		cache->user_parts[i].config = &part_config[i];
 		cache->user_parts[i].runtime = &part_runtime[i];
+		env_spinlock_init(&cache->user_parts[i].runtime->reserved_clines_barrier);
```
arutk, line 566: it' suprisning for me to see a spin lock variable named "barrier". How about reserved_clines_lock?

arutk (changes requested):

tests/unit/tests/utils/utils_part.c/utils_part.c
```diff
@@ -0,0 +1,473 @@
...
+{
+	return mock();
+}
+
+ocf_cache_line_t __wrap_ocf_freelist_num_free(ocf_freelist_t freelist)
+{
+	return mock();
+}
+
+static void _how_much_to_evict_test01(void **state)
```
arutk, line 61: Tests look really comprehensive, I really like it. Please consider adding one for req->part_evict flag - might be convenient to add getter/setter functions for easy mocking.

src/utils/utils_part.h
```diff
@@ -61,7 +80,37 @@ static inline void ocf_part_sort(struct ocf_cache *cache)
 	ocf_lst_sort(&cache->lst_part);
 }
 
-static inline ocf_cache_mode_t ocf_part_get_cache_mode(struct ocf_cache *cache,
+static inline bool _is_occupancy_for_ioclass_enabled(struct ocf_request *req)
```
arutk, line 83: suggest renaming to "ocf_ioclass_ocupancy_enabled"

arutk: "ocf_part_is_enabled" fits into the context even better

src/engine/engine_common.c
```diff
@@ -432,13 +427,172 @@ static int lock_clines(struct ocf_request *req,
...
+
+	*lock_status = lock_clines(req, engine_cbs);
+
+unlock:
+	ocf_req_hash_unlock_wr(req);
+
+	return res;
+}
+
+static inline int _evict_and_map(struct ocf_request *req,
```
arutk, line 549: suggest to rename to ocf_allot_evict

src/engine/engine_common.c
```diff
@@ -432,13 +427,172 @@ static int lock_clines(struct ocf_request *req,
...
+}
+
+static inline int _handle_mapping(struct ocf_request *req,
+		const struct ocf_engine_callbacks *engine_cbs)
+{
+	int lock_status = -ENOENT;
+	uint32_t clines_to_evict = 0;
+	int res;
+
+	res = _reserve_and_map(req, engine_cbs, &clines_to_evict, &lock_status);
```
arutk, line 582: suggest to rename function to ocf_allot_free

src/engine/engine_common.c
```diff
@@ -432,13 +427,172 @@ static int lock_clines(struct ocf_request *req,
...
+		/* Some cache lines still belong to particualr ioclass. Evict small
+		 * portion of them before further proccesing.  */
+		req->part_evict = true;
+		ocf_metadata_start_exclusive_access(metadata_lock);
+		space_managment_evict_do(req->cache, req, 256);
+		ocf_metadata_end_exclusive_access(metadata_lock);
+	}
+}
+
+static inline int evict_and_reserve(struct ocf_request *req,
```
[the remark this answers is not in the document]
arutk: correction: I suggest ocf_evict_part_overflow

src/metadata/metadata_partition_structs.h
```diff
@@ -11,33 +11,38 @@
...
 struct ocf_user_part_runtime {
-        uint32_t curr_size;
-        uint32_t head;
-        struct eviction_policy eviction;
-        struct cleaning_policy cleaning;
+	env_atomic curr_size;
+
+	/* Number of cache lines reserved for in flight IO */
+	uint32_t reserved_clines_cnt;
+	env_spinlock reserved_clines_lock;
```
arutk, line 34: This is part of persistent metadata, written to disk without any serialization upon management operation. env_atomic and env_spinlock are opaque types, potentially containing pointers. For this reason you should not put atomics/spinlocks in ocf_user_part_runtime. Please either change the type (curr_size back to regular type) or move these members outside this struct. 

Also consider what actually belongs to persistent metadata - spinlock obviously does not, while curr_size probably does (depende on usage details).

src/mngt/ocf_mngt_cache.c
```diff
@@ -172,7 +172,7 @@ static void __init_partitions(ocf_cache_t cache)
 
 	/* Init default Partition */
 	ENV_BUG_ON(ocf_mngt_add_partition_to_cache(cache, PARTITION_DEFAULT,
-			"unclassified", 0, PARTITION_SIZE_MAX,
+			"unclassified", 0, cache->conf_meta->cachelines,
```
arutk, line 175: This line is part of cache start, performed before cache attach. You cannot depend on conf_meta->cachelines being correctly set here. That's the reason why PARTITION_SIZE_MAX was used. You would probably need to iterate over partitions and trim max size to conf_meta-cachelines during cache attach

Alternatively give up setting partition max size to cache size and allow partition max to be larger than cache size - see my next comment

src/mngt/ocf_mngt_cache.c
```diff
@@ -193,10 +193,13 @@ static void __init_partitions_attached(ocf_cache_t cache)
 {
 	ocf_part_id_t part_id;
 
+	ocf_mngt_set_partition_size(cache, PARTITION_DEFAULT, 0,
```
arutk, line 196: looks like this is the place where you have conf_meta->cachelines ready

src/mngt/ocf_mngt_io_class.c
```diff
@@ -41,7 +41,7 @@ int ocf_mngt_add_partition_to_cache(struct ocf_cache *cache,
 	if (cache->user_parts[part_id].config->flags.valid)
 		return -OCF_ERR_INVAL;
 
-	if (max_size > PARTITION_SIZE_MAX)
+	if (max_size > cache->conf_meta->cachelines)
```
arutk, line 44: consider allowing partition max size > cache size. it is tricky to enforce actual limit (don't know actual cache size before attach) and partition max size > cache size makes sense assuming user can detach cache drive and attach larger one.

src/metadata/metadata_partition_structs.h
```diff
@@ -11,33 +11,38 @@
...
+			*/
+	} flags;
 };
 
 struct ocf_user_part_runtime {
-        uint32_t curr_size;
-        uint32_t head;
-        struct eviction_policy eviction;
-        struct cleaning_policy cleaning;
+	env_atomic curr_size;
```
arutk, line 30: why do you need curr_size to be atomic?

src/utils/utils_part.c
```diff
@@ -184,9 +184,131 @@ void ocf_part_set_valid(struct ocf_cache *cache, ocf_part_id_t id,
...
 			part->config->min_size = 0;
-			part->config->max_size = PARTITION_SIZE_MAX;
+			part->config->max_size = cache->conf_meta->cachelines;
 			ENV_BUG_ON(env_strncpy(part->config->name, sizeof(part->config->name),
 					"Inactive", 9));
 		}
 	}
 }
+
+static inline uint32_t _how_much_to_evict(struct ocf_request *req)
```
arutk, line 194: how about changing name to "ocf_part_evict_size" ?

src/utils/utils_part.c
```diff
@@ -184,9 +184,131 @@ void ocf_part_set_valid(struct ocf_cache *cache, ocf_part_id_t id,
...
+		/* Enough free cachelines available and they can be assigned to target
+		 * partition */
+		cache_lines_to_evict = 0;
+
+	}
+
+	return cache_lines_to_evict + occupancy_debt;
+}
+
+static inline void _reserve_cache_lines(struct ocf_request *req)
```
arutk, line 247: rename to ocf_part_reserve_cache_lines?

arutk (changes requested):

src/engine/engine_common.c
```diff
@@ -432,13 +425,149 @@ static int lock_clines(struct ocf_request *req,
 	}
 }
 
+static inline int ocf_evict_part_overflow(struct ocf_request *req,
+		const struct ocf_engine_callbacks *engine_cbs, uint32_t clines_to_evict)
+{
+	int lock_status = -ENOENT;
```
imjfckm, line 431: Why errno instead of OCF error codes?

src/utils/utils_part.c
```diff
@@ -184,7 +184,7 @@ void ocf_part_set_valid(struct ocf_cache *cache, ocf_part_id_t id,
 			cache->conf_meta->valid_parts_no--;
 			part->config->priority = OCF_IO_CLASS_PRIO_LOWEST;
 			part->config->min_size = 0;
-			part->config->max_size = PARTITION_SIZE_MAX;
+			part->config->max_size = cache->conf_meta->cachelines;
```
arutk, line 187: I think it still isn't correct. If I recall correctly IO classes can be configured with cache device detached. Please add a simple functional test to check how max io class occupancy behaves when configured with cache detached and then what happens after attaching a cache device.

src/mngt/ocf_mngt_io_class.c
```diff
@@ -85,8 +85,8 @@ int ocf_mngt_set_partition_size(struct ocf_cache *cache,
 		return -OCF_ERR_INVAL;
 	}
 
-	if (max > PARTITION_SIZE_MAX)
-		max = PARTITION_SIZE_MAX;
```
arutk, line 89: same potential problem with detached cache device here

src/ocf_stats.c
```diff
@@ -304,8 +306,9 @@ int ocf_core_io_class_get_stats(ocf_core_t core, ocf_part_id_t part_id,
 	stats->dirty_clines = env_atomic_read(&core->runtime_meta->
 			part_counters[part_id].dirty_clines);
 
-	stats->free_clines = cache->conf_meta->cachelines -
-			cache_occupancy_total;
+	max_size = cache->user_parts[part_id].config->max_size;
+	curr_size = ocf_part_get_occupancy(cache, part_id);
+	stats->free_clines = (max_size >= curr_size) ? max_size - curr_size : 0;
```
arutk, line 311: so "free cachelines" now means how far an IO class is from reaching its max? even if cache occupancy is 100%? I think neither approach makes sense...  I think we should only report occupancy for I/O class, "free" counter is meaningless. What do you think?

src/engine/engine_common.c
```diff
@@ -432,13 +425,149 @@ static int lock_clines(struct ocf_request *req,
...
+		const struct ocf_engine_callbacks *engine_cbs)
+{
+	int lock_status = -ENOENT;
+	uint32_t clines_to_evict;
+	int res;
+
+	/* Cachelines are mapped in correct partition */
+	if (!ocf_engine_needs_repart(req)) {
+		ocf_req_hash_unlock_rd(req);
+		return lock_clines(req, engine_cbs);
```
imjfckm, line 462: Shouldn't we lock clines before unlocking hash buckets?

src/utils/utils_part.c
```diff
@@ -184,9 +184,85 @@ void ocf_part_set_valid(struct ocf_cache *cache, ocf_part_id_t id,
...
+		/* Number of cachelines to insert greater than number of free
+		 * cachelines */
+		if (part_available >= needed_cache_lines) {
+			/* Cache is full, but target's part occupancy limit is not reached
+			   */
+			ocf_req_reset_part_evict(req);
+			cache_lines_to_evict = needed_cache_lines;
+		} else {
+			/* Cache is full and target part reached it's occupancy limit */
+			ocf_req_set_part_evict(req);
```
imjfckm, line 227: Why use getter/setter here (not inlined) if other `struct ocf_request` fields are accessed by pointer dereference?

src/utils/utils_part.h
```diff
@@ -73,7 +74,27 @@ static inline void ocf_part_sort(struct ocf_cache *cache)
 	ocf_lst_sort(&cache->lst_part);
 }
 
-static inline ocf_cache_mode_t ocf_part_get_cache_mode(struct ocf_cache *cache,
+static inline bool ocf_part_is_enabled(struct ocf_request *req)
+{
+	return req->cache->user_parts[req->part_id].config->max_size != 0;
+}
+
+#define OCF_PART_SUCCESS	0
```
arutk, line 82: How about renaming these constants to OCF_ALLOT_(SUCCESS/....) ?

src/engine/engine_common.c
arutk: Yes, we should

arutk (changes requested):

src/utils/utils_part.c
mmichal10: I've added set&get methods to make this function more unit-testable

src/utils/utils_part.h
mmichal10: In current implementation no allocation takes place. How about OCF_PART_HAS_SPACE and OCF_PART_IS_FULL?

src/eviction/eviction.c
```diff
@@ -119,29 +119,30 @@ static inline uint32_t ocf_evict_do(ocf_cache_t cache,
 	return evicted;
 }
 
-int space_managment_evict_do(struct ocf_cache *cache,
-		struct ocf_request *req, uint32_t evict_cline_no)
```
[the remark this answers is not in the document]
mmichal10: It were leftovers from previous version

src/utils/utils_part.c
imjfckm: Well.. It's kinda inconsistent with other fields usages, but OK. Could you make them `inline`?

src/utils/utils_part.h
arutk: looks good, thanks :)

src/engine/engine_common.c
```diff
@@ -453,49 +585,19 @@ int ocf_engine_prepare_clines(struct ocf_request *req,
...
 	mapped = ocf_engine_is_mapped(req);
 	if (mapped) {
-		/* Request cachelines are already mapped, acquire cacheline
-		 * lock */
-		lock = lock_clines(req, engine_cbs);
+		return ocf_prepare_clines_hit(req, engine_cbs);
 	} else {
 		/* check if request should promote cachelines */
 		promote = ocf_promotion_req_should_promote(
 				req->cache->promotion_policy, req);
```
arutk, line 592: you can reduce indentation level here after early return in line 588

src/engine/engine_common.c
```diff
@@ -432,13 +425,152 @@ static int lock_clines(struct ocf_request *req,
...
+
+static inline int ocf_prepare_clines_hit(struct ocf_request *req,
+		const struct ocf_engine_callbacks *engine_cbs)
+{
+	int lock_status = -OCF_ERR_NO_LOCK;
+	uint32_t clines_to_evict;
+	int res;
+
+	/* Cachelines are mapped in correct partition */
+	if (!ocf_engine_needs_repart(req)) {
```
arutk, line 462: what if target partition is disabled? shouldn't we attempt to evict from it here?

src/utils/utils_part.h
```diff
@@ -61,7 +107,27 @@ static inline void ocf_part_sort(struct ocf_cache *cache)
 	ocf_lst_sort(&cache->lst_part);
 }
 
-static inline ocf_cache_mode_t ocf_part_get_cache_mode(struct ocf_cache *cache,
+static inline bool ocf_part_is_enabled(struct ocf_request *req)
```
arutk, line 110: I suggest to change function parameter to part_id

src/utils/utils_part.h
```diff
@@ -50,6 +51,51 @@ static inline ocf_part_id_t ocf_part_class2id(ocf_cache_t cache, uint64_t class)
...
+static inline uint32_t ocf_part_get_min_size(ocf_cache_t cache,
+		ocf_part_id_t part_id)
+{
+	uint64_t ioclass_size;
+
+	if (unlikely(cache->user_parts[part_id].config->min_size > 100)) {
+		ocf_cache_log(cache, log_err, "Ioclass %hu: invalid min size %u\n",
+				part_id, cache->user_parts[part_id].config->min_size);
+		ENV_BUG();
+	}
```
arutk, line 69: I would expect to see this check wherever we set the min value rather than here in get_min_size()

src/utils/utils_part.h
```diff
@@ -50,6 +51,51 @@ static inline ocf_part_id_t ocf_part_class2id(ocf_cache_t cache, uint64_t class)
...
+	if (unlikely(cache->user_parts[part_id].config->max_size > 100)) {
+		ocf_cache_log(cache, log_err, "Ioclass %hu: invalid max size %u\n",
+				part_id, cache->user_parts[part_id].config->max_size);
+		ENV_BUG();
+	}
+
+	ioclass_size = cache->user_parts[part_id].config->max_size *
+		cache->conf_meta->cachelines;
+
+	ioclass_size /= 100;
```
arutk, line 94: per our discussion: would be better to round up here

src/engine/engine_common.c
```diff
@@ -432,13 +425,152 @@ static int lock_clines(struct ocf_request *req,
...
+	if (res != OCF_PART_IS_FULL)
+		return lock_status;
+
+	/* Target partition is full. Evict missing number of cache lines */
+	lock_status = ocf_evict_part_overflow(req, engine_cbs, clines_to_evict);
+
+	return lock_status;
+}
+
+static inline int ocf_map_free(struct ocf_request *req,
```
arutk, line 487: I fell that the way code is split between subroutines and ocf_engine_prepare_cachelines is artificial. For example this function assumes it is called with hash bucket read lock acquired and it exits with locks released. Seems to me that one long function would be easier to follow. Let's discuss how/if to introduce new sub-routines.

arutk (changes requested):

src/utils/utils_part.c
```diff
@@ -190,3 +190,79 @@ void ocf_part_set_valid(struct ocf_cache *cache, ocf_part_id_t id,
...
+	}
+
+	if (ocf_freelist_num_free(req->cache->freelist) <
+			ocf_engine_unmapped_count(req)) {
+		/* Number of cachelines to insert greater than number of free
+		 * cachelines */
+		if (part_available >= needed_cache_lines) {
+			/* Cache is full, but target's part occupancy limit is not reached
+			   */
+			ocf_req_reset_part_evict(req);
```
arutk, line 223: I personally find "unset" or "clear" more intuitive than "reset" - please consider changing

src/utils/utils_part.c
```diff
@@ -190,3 +190,79 @@ void ocf_part_set_valid(struct ocf_cache *cache, ocf_part_id_t id,
...
+		ocf_req_set_part_evict(req);
+
+	} else if (part_available >= needed_cache_lines) {
+		/* Enough free cachelines available and they can be assigned to target
+		 * partition */
+		cache_lines_to_evict = 0;
+
+	}
+
+	return cache_lines_to_evict + part_occupancy_debt;
```
arutk, line 244: Maybe we should consider limiting size of eviction induced by occupancy debt to some fixed constant? This would limit the overhead of occupancy debt on a single I/O

src/engine/engine_common.c
```diff
@@ -432,13 +425,159 @@ static int lock_clines(struct ocf_request *req,
...
+		goto unlock;
+	}
+
+	lock_status = lock_clines(req, engine_cbs);
+
+unlock:
+	ocf_metadata_end_exclusive_access(metadata_lock);
+
+	return lock_status;
+}
```
arutk, line 452: I think you meant to remove this function

src/engine/engine_common.c
```diff
@@ -432,13 +425,159 @@ static int lock_clines(struct ocf_request *req,
...
+    /* Check whether target partition if full */
+	if (res == OCF_PART_HAS_SPACE)
+		lock_status = lock_clines(req, engine_cbs);
+
+	if (res == OCF_PART_IS_DISABLED)
+		ocf_req_set_mapping_error(req);
+
+	ocf_req_hash_unlock_rd(req);
+
+	if (res != OCF_PART_IS_FULL)
```
arutk, line 481: if (res == OCF_PART_HAS_SPACE)

src/engine/engine_common.c
```diff
@@ -432,13 +425,159 @@ static int lock_clines(struct ocf_request *req,
...
+	}
+
+	res = ocf_part_check_space(req, &clines_to_evict);
+
+    /* Check whether target partition if full */
+	if (res == OCF_PART_HAS_SPACE)
+		lock_status = lock_clines(req, engine_cbs);
+
+	if (res == OCF_PART_IS_DISABLED)
+		ocf_req_set_mapping_error(req);
```
arutk, line 477: this is premature, it is done again (correctly) in line 492

src/engine/engine_common.c
```diff
@@ -432,13 +425,159 @@ static int lock_clines(struct ocf_request *req,
...
+
+		ocf_req_hash_unlock_wr(req);
+		return lock_status;
+	}
+
+eviction:
+	ocf_req_hash_unlock_wr(req);
+	ocf_metadata_start_exclusive_access(metadata_lock);
+
+	if (space_managment_evict_do(req->cache, req, clines_to_evict) ==
```
arutk, line 546: clines_to_evict should be re-calculated after releasing metadata lock

src/engine/engine_common.c
```diff
@@ -432,13 +425,159 @@ static int lock_clines(struct ocf_request *req,
...
+	int res;
+
+	/* Mapping must be performed holding (at least) hash-bucket write lock */
+	ocf_req_hash_lock_upgrade(req);
+
+	/* Verify whether partition occupancy threshold is not reached yet or cache
+	 * is not out of free cachelines */
+	res = ocf_part_check_space(req, &clines_to_evict);
+	if (res == OCF_PART_IS_DISABLED) {
+		ocf_req_set_mapping_error(req);
```
arutk, line 520: wouldn't we want to evict something in case of OCF_PART_IS_DISABLED?

arutk (changes requested):

src/engine/engine_common.c
mmichal10: This check is correct - if part would be disabled, but had cachelines assigned, `ocf_part_check_space()` return `OCF_PART_IS_FULL`. Any other status means that no eviction is needed. It also implies correctness of mapping error set in :477. I will add comment to my code

src/engine/engine_common.c
mmichal10: Eviction is needed only if status is OCF_PART_IS_FULL

src/mngt/ocf_mngt_io_class.c
```diff
@@ -51,6 +56,14 @@ int ocf_mngt_add_partition_to_cache(struct ocf_cache *cache,
...
 	}
 
+	for_each_lst(&cache->lst_part, iter, iter_id) {
+		if (iter_id == part_id) {
+			ocf_cache_log(cache, log_err,
+					"Part with id %hu already exists\n", part_id);
+			return -OCF_ERR_INVAL;
+		}
+	}
+
```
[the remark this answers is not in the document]
mmichal10: Re-configuration is done via `ocf_mngt_cache_io_classes_configure()`. `ocf_mngt_add_partition_to_cache()` is a routine called on cache start and calling it with the same id twice leads to bug in `ocf_part_sort()`

src/ocf_stats_builder.c
```diff
@@ -229,15 +229,9 @@ static void _ocf_stats_part_fill(ocf_cache_t cache, ocf_part_id_t part_id,
...
-		if (part_id == PARTITION_DEFAULT) {
-			_set(&usage->free,
-				_lines4k(stats->free_clines, cache_line_size),
-				_lines4k(cache_size, cache_line_size));
-		} else {
-			_set(&usage->free,
-				_lines4k(0, cache_line_size),
-				_lines4k(0, cache_line_size));
-		}
+		_set(&usage->free,
```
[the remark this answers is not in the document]
mmichal10: Since `free_clines` field is irrelevant, I would leave setting it's value to zero. I believe we should let adapter and cli utility decide whether print this information or not.

src/mngt/ocf_mngt_io_class.c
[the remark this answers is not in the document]
arutk: i see, thanks

src/engine/engine_common.c
arutk: got it

src/engine/engine_common.c
arutk: I think this does not apply any more (since you rejected my previous comment)

src/engine/engine_common.c
arutk: ok

src/utils/utils_part.c
```diff
@@ -190,3 +190,80 @@ void ocf_part_set_valid(struct ocf_cache *cache, ocf_part_id_t id,
...
+		ocf_req_set_part_evict(req);
+
+	} else if (part_available >= needed_cache_lines) {
+		/* Enough free cachelines available and they can be assigned to target
+		 * partition */
+		cache_lines_to_evict = 0;
+
+	}
+
+	return min((uint32_t)OCF_PENDING_EVICTION_LIMIT,
```
arutk, line 244: We should only limit 'part_occupancy_debt'. Otherwise we risk evicting to few cachelines in case of regular miss for a large requrest

src/utils/utils_part.h
```diff
@@ -50,6 +51,39 @@ static inline ocf_part_id_t ocf_part_class2id(ocf_cache_t cache, uint64_t class)
...
+}
+
+
+static inline uint32_t ocf_part_get_max_size(ocf_cache_t cache,
+		ocf_part_id_t part_id)
+{
+	uint64_t ioclass_size;
+
+	ioclass_size = cache->user_parts[part_id].config->max_size *
+		cache->conf_meta->cachelines;
```
arutk, line 80: aren both operands 32-bit? If so, there is a potential for overflow here despite assigning to wider type.

arutk: please make sure max_size = 100 allows you to fill entire cache with cache device large enough to contain more than MAX_UINT32 / 100. This should be easy on null block device

arutk (approved):

src/utils/utils_part.h
mmichal10: I did the test, and it works fine

arutk (approved):

Ostrokrzew (commented):
Looks fine, but didn't check tests.

arutk (approved):

robertbaldyga (approved):

PR #422, opened 2020-11-24
reduce metadata interface overhead

arutk:
1. Removing metadata abstractions:
  - main metadata abstraction  (the one with hash implementation)
  - metadata bits cacheline size abstraction
  - collision layout (seq vs striping) abstraction.
2. Slight refactoring of metadata segment ops and superblock code in order to reduce coupling.


src/metadata/metadata_superblock.c
```diff
@@ -0,0 +1,431 @@
...
+	}
+
+	ocf_pipeline_next(pipeline);
+}
+
+static void ocf_metadata_check_crc_if_clean(ocf_pipeline_t pipeline,
+		void *priv, ocf_pipeline_arg_t arg)
+{
+	ocf_metadata_check_crc_skip(pipeline, priv, arg, true);
+}
```
robertbaldyga, line 122: This could be moved to `metadata_segment_ops.c`. This function is the only reason for exporting `ocf_metadata_check_crc_skip()`, otherwise it could be made static.

robertbaldyga: Btw calculating CRC is dependent on superblock segment and this split between superblock/segment ops/other segments seems to lack ability to reflect this dependency gracefully. Maybe we can find a way to make ops superblock agnostic - e.g. inject dependency to superblock by some abstraction or make other sections superblock aware so that the coupling isn't introduced by ops themselves?

robertbaldyga (changes requested):

src/metadata/metadata_superblock.c
arutk: done

arutk:
retest this please

robertbaldyga (approved):

mmichal10 (approved):

PR #426, opened 2020-12-07
Remove memcpy from collision/eviction policy metadata api

arutk:

### DIFF
--- a/src/cleaning/alru.c
+++ b/src/cleaning/alru.c
@@ -98,7 +98,7 @@ static void add_alru_head(struct ocf_cache *cache, int partition_id,
 	unsigned int curr_head_index;
 	unsigned int collision_table_entries = cache->device->collision_table_entries;
 	struct ocf_user_part *part = &cache->user_parts[partition_id];
-	struct cleaning_policy_meta policy;
+	struct alru_cleaning_policy_meta *alru;
 
 	ENV_BUG_ON(!(collision_index < collision_table_entries));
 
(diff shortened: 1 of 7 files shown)


src/cleaning/alru.c
```diff
@@ -727,19 +703,19 @@ static int get_data_to_flush(struct alru_context *alru)
...
 
 		cache_line = part->runtime->cleaning.policy.alru.lru_tail;
 
 		last_access = compute_timestamp(config);
 
 		OCF_DEBUG_PARAM(cache, "Last access=%u, timestamp=%u rel=%d",
-				last_access, policy.meta.alru.timestamp,
-				policy.meta.alru.timestamp < last_access);
+				last_access, alru->timestamp,
+				alru->timestamp < last_access);
```
robertbaldyga, line 714: `alru` is uninitialized at this point.

robertbaldyga (changes requested):

src/cleaning/alru.c
arutk: good catch, thanks

robertbaldyga (approved):

src/cleaning/alru.c
```diff
@@ -466,31 +442,31 @@ int cleaning_policy_alru_initialize(ocf_cache_t cache, int init_metadata)
...
 	struct ocf_user_part *part;
 	ocf_part_id_t part_id;
-	struct alru_context *alru;
+	struct alru_context *ctx;
 	int error = 0;
 	unsigned i;
 
-	alru = env_vzalloc(sizeof(*alru));
-	if (!alru) {
+	ctx = env_vzalloc(sizeof(*ctx));
```
Ostrokrzew, line 449: I have a question. Don't we want to remove vmallocs '& Co' from CAS? Just asking because we had this topic long ago.

arutk: Our biggest concern is using vmallocs in performance sensitive paths (like I/O service). This is a management path, so vmalloc looks like a good idea here. It is slower, but does not require memory pages to be physically continuous like kmalloc. And we don't need physically continuous pages here, since this structure is referenced from CPU only (no DMA / hardware access).

Ostrokrzew: Thanks for the explanation.

Ostrokrzew (approved):
Nice. I like your style.

PR #428, opened 2020-12-09
[RFC] cacheline concurrency changes for parallel eviction

arutk:


src/engine/cache_engine.h
```diff
@@ -12,6 +12,7 @@ struct ocf_request;
 #define LOOKUP_HIT 5
 #define LOOKUP_MISS 6
 #define LOOKUP_INSERTED 8
+#define LOOKUP_REMAPPED 9
```
mmichal10, line 15: You mentioned `LOOKUP_REPLACED` in commit message

src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -756,6 +764,11 @@ static int _ocf_req_trylock_rd(struct ocf_request *req)
 	if (ret == OCF_LOCK_NOT_ACQUIRED) {
 		/* Request is not locked, discard acquired locks */
 		for (; i >= 0; i--) {
+			if (!_ocf_req_needs_cl_lock(req, i)) {
+				/* nothing to lock */
```
mmichal10, line 768: Shouldn't the comment be sth like `no lock to discard`?

src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -815,6 +828,9 @@ static int _ocf_req_lock_rd(struct ocf_request *req, ocf_req_async_lock_cb cb)
 
 err:
 	for (; i >= 0; i--) {
+		if (req->map[i].status <= LOOKUP_MISS) {
```
mmichal10, line 831: Don't you think `req->map[i].status == LOOKUP_MISS || req->map[i].status == LOOKUP_HIT` would be more readable?

arutk: this is an error, I meant to use 
if (!_ocf_req_needs_cl_lock(req, i))

PR #430, opened 2020-12-16
Create separate pipelines and paths for cache attach/load scenarios

rafalste:


src/mngt/ocf_mngt_cache.c
```diff
@@ -1542,13 +1609,15 @@ struct ocf_pipeline_properties _ocf_mngt_cache_attach_pipeline_properties = {
 	.priv_size = sizeof(struct ocf_cache_attach_context),
 	.finish = _ocf_mngt_cache_attach_finish,
 	.steps = {
+		OCF_PL_STEP(_ocf_mngt_copy_uuid_data),
 		OCF_PL_STEP(_ocf_mngt_attach_cache_device),
-		OCF_PL_STEP(_ocf_mngt_attach_load_properties),
+		OCF_PL_STEP(_ocf_mngt_init_properties),
+		OCF_PL_STEP(_ocf_mngt_attach_properties),
```
arutk, line 1615: load in "_ocf_mngt_load_properties" reffers to loading properties rather than OCF cache load. To avoid ambiguity I would rename "load" to "read", so here we would have "_ocf_mng_attach_read_properties".

src/mngt/ocf_mngt_cache.c
```diff
@@ -1558,6 +1627,29 @@ struct ocf_pipeline_properties _ocf_mngt_cache_attach_pipeline_properties = {
...
 };
 
+struct ocf_pipeline_properties _ocf_mngt_cache_load_pipeline_properties = {
+	.priv_size = sizeof(struct ocf_cache_attach_context),
+	.finish = _ocf_mngt_cache_attach_finish,
+	.steps = {
+		OCF_PL_STEP(_ocf_mngt_copy_uuid_data),
+		OCF_PL_STEP(_ocf_mngt_attach_cache_device),
+		OCF_PL_STEP(_ocf_mngt_init_properties),
+		OCF_PL_STEP(_ocf_mngt_load_properties),
```
arutk, line 1637: load in "_ocf_mngt_load_properties" reffers to loading properties rather than OCF cache load. To avoid ambiguity I would rename "load" to "read", so here we would have "_ocf_mng_load_read_properties".

src/mngt/ocf_mngt_cache.c
```diff
@@ -938,16 +935,38 @@ static void _ocf_mngt_attach_load_properties_end(void *priv, int error,
 	context->metadata.shutdown_status = properties->shutdown_status;
 	context->metadata.dirty_flushed = properties->dirty_flushed;
```
arutk, line 936: Looks like these doesn't make any sense in attach. Can you please verify this? We can talk about this face to face

src/mngt/ocf_mngt_cache.c
```diff
@@ -911,18 +912,14 @@ static void _ocf_mngt_attach_load_properties_end(void *priv, int error,
 
 	if (error) {
```
arutk, line 913: if (error == -OCF_ERR_NO_METADATA)

src/mngt/ocf_mngt_cache.c
```diff
@@ -911,18 +912,14 @@ static void _ocf_mngt_attach_load_properties_end(void *priv, int error,
...
+		 * dismiss error and go to the next step in the pipeline.
 		 */
-		if (error == -OCF_ERR_NO_METADATA &&
-				cache->device->init_mode != ocf_init_mode_load)
-			OCF_PL_NEXT_RET(context->pipeline);
-		else
-			OCF_PL_FINISH_RET(context->pipeline, error);
-	} else if (cache->device->init_mode != ocf_init_mode_load) {
+		OCF_PL_NEXT_RET(context->pipeline);
+	} else {
```
arutk, line 919: if (!error), without "else"

src/mngt/ocf_mngt_cache.c
```diff
@@ -938,16 +935,38 @@ static void _ocf_mngt_attach_load_properties_end(void *priv, int error,
 	context->metadata.shutdown_status = properties->shutdown_status;
 	context->metadata.dirty_flushed = properties->dirty_flushed;
 
-	if (cache->device->init_mode == ocf_init_mode_load) {
-		context->metadata.line_size = properties->line_size;
-		cache->conf_meta->metadata_layout = properties->layout;
-		cache->conf_meta->cache_mode = properties->cache_mode;
+	ocf_pipeline_next(context->pipeline);
```
arutk, line 938: actually looks like this unused in current PR revision. both OCF_PL_FINISH_RET and OCF_PL_NEXT_RET do "return"

src/mngt/ocf_mngt_cache.c
```diff
@@ -1558,6 +1589,28 @@ struct ocf_pipeline_properties _ocf_mngt_cache_attach_pipeline_properties = {
...
+		OCF_PL_STEP(_ocf_mngt_attach_cache_device),
+		OCF_PL_STEP(_ocf_mngt_init_properties),
+		OCF_PL_STEP(_ocf_mngt_load_read_properties),
+		OCF_PL_STEP(_ocf_mngt_attach_check_ram),
+		OCF_PL_STEP(_ocf_mngt_load_check_metadata),
+		OCF_PL_STEP(_ocf_mngt_attach_prepare_metadata),
+		OCF_PL_STEP(_ocf_mngt_test_volume),
+		OCF_PL_STEP(_ocf_mngt_load_superblock),
+		OCF_PL_STEP(_ocf_mngt_attach_instance),
+		OCF_PL_STEP(_ocf_mngt_init_instance_load),
```
arutk, line 1606: "_ocf_mngt_load_init_instance" would be more in line with our convention.

src/mngt/ocf_mngt_cache.c
```diff
@@ -1340,21 +1385,18 @@ static void _ocf_mngt_attach_load_superblock_complete(void *priv, int error)
...
-		OCF_PL_NEXT_RET(context->pipeline);
-
 	ocf_cache_log(cache, log_info, "Loading cache state...\n");
 	ocf_metadata_load_superblock(cache,
-			_ocf_mngt_attach_load_superblock_complete, context);
+			_ocf_mngt_load_superblock_complete, context);
 }
 
-static void _ocf_mngt_attach_init_instance(ocf_pipeline_t pipeline,
+static void _ocf_mngt_attach_instance(ocf_pipeline_t pipeline,
```
arutk, line 1399: This name is not very informative. I think this function actually does two things: initializes cleaning policy and cleaner. So maybe splitting it into two pipeline steps would be the cleanest solution:
_ocf_mngt_init_promotion
_ocf_mngt_init_cleaner

arutk (changes requested):

src/mngt/ocf_mngt_cache.c
```diff
@@ -369,7 +366,7 @@ static void _ocf_mngt_close_all_uninitialized_cores(
  * @brief routine loading metadata from cache device
  *  - attempts to open all the underlying cores
  */
-static int _ocf_mngt_init_instance_add_cores(
+static int _ocf_mngt_load_init_instance_add_cores(
```
arutk, line 369: I suggest _ocf_mngt_load_add_cores  (if there are no name conflicts)

src/mngt/ocf_mngt_cache.c
```diff
@@ -960,14 +957,44 @@ static void _ocf_mngt_attach_load_properties(ocf_pipeline_t pipeline,
...
+{
+	struct ocf_cache_attach_context *context = priv;
+	ocf_cache_t cache = context->cache;
+
 	if (context->cfg.force)
-		OCF_PL_NEXT_RET(context->pipeline);
+		OCF_PL_NEXT_RET(pipeline);
+
+	if (cache->metadata.is_volatile)
+		OCF_PL_NEXT_RET(pipeline);
```
arutk, line 973: We must check for pre-existing metadata even if starting in volatile mode. In volatile mode we will not write our own superblock to the device, but we should check if there already is one.

src/mngt/ocf_mngt_cache.c
```diff
@@ -960,14 +957,44 @@ static void _ocf_mngt_attach_load_properties(ocf_pipeline_t pipeline,
...
+			_ocf_mngt_load_read_properties_end, context);
+}
+
+static void _ocf_mngt_load_check_metadata(ocf_pipeline_t pipeline,
+		void *priv, ocf_pipeline_arg_t arg)
+{
+	struct ocf_cache_attach_context *context = priv;
+
+	if (context->metadata.status)
+		OCF_PL_FINISH_RET(pipeline, -OCF_ERR_START_CACHE_FAIL);
```
arutk, line 995: since we are reworking this area anyway: maybe we can do better then -OCF_ERR_START_CACHE_FAIL? I suppose context->metadata.status already contains meaningfull error (maybe good to verify this).

src/mngt/ocf_mngt_cache.c
```diff
@@ -991,62 +1013,44 @@ static void _ocf_mngt_attach_prepare_metadata(ocf_pipeline_t pipeline,
 	if (ocf_metadata_init_variable_size(cache, context->volume_size,
 			context->metadata.line_size,
 			cache->conf_meta->metadata_layout)) {
-		OCF_PL_FINISH_RET(context->pipeline, -OCF_ERR_START_CACHE_FAIL);
+		OCF_PL_FINISH_RET(pipeline, -OCF_ERR_START_CACHE_FAIL);
```
arutk, line 1016: please try to return better status

src/mngt/ocf_mngt_cache.c
```diff
@@ -991,62 +1013,44 @@ static void _ocf_mngt_attach_prepare_metadata(ocf_pipeline_t pipeline,
...
 	}
 	context->flags.attached_metadata_inited = true;
 
 	cache->freelist = ocf_freelist_init(cache);
 	if (!cache->freelist)
-		OCF_PL_FINISH_RET(context->pipeline, -OCF_ERR_START_CACHE_FAIL);
+		OCF_PL_FINISH_RET(pipeline, -OCF_ERR_START_CACHE_FAIL);
```
arutk, line 1022: please try to return better status

src/mngt/ocf_mngt_cache.c
```diff
@@ -2097,19 +2160,11 @@ void ocf_mngt_cache_load(ocf_cache_t cache,
 	if (cache->metadata.is_volatile)
 		OCF_CMPL_RET(cache, priv, -OCF_ERR_INVAL);
 
-	/* Load is not allowed with 'force' flag on */
-	if (cfg->force) {
-		ocf_cache_log(cache, log_err, "Using 'force' flag is forbidden "
-				"for load operation.");
-		OCF_CMPL_RET(cache, priv, -OCF_ERR_INVAL);
-	}
-
```
arutk, line 2106: shouldn't this be kept?

arutk: Also please make sure a comment for cfg->flag in API header file is up to date

arutk (changes requested):

src/mngt/ocf_mngt_cache.c
```diff
@@ -1290,6 +1273,43 @@ static void _ocf_mngt_init_attached_nonpersistent(ocf_cache_t cache)
...
+	OCF_PL_NEXT_RET(pipeline);
+
+err_uuid:
+	env_vfree(data);
+err_pipeline:
+	ocf_pipeline_destroy(pipeline);
+	ocf_pipeline_destroy(cache->stop_pipeline);
+	OCF_CMPL_RET(cache, priv1, priv2, result);
+}
+
```
arutk, line 1312: you should call OCF_PL_FINISH_RET(pipeline, result) instead of deallocating pipeline manually - this made sense when the code was executed outside of pipeline context

src/ocf_freelist.c
```diff
@@ -397,7 +397,8 @@ ocf_freelist_t ocf_freelist_init(struct ocf_cache *cache)
 		env_atomic64_set(&freelist->part[i].curr_size, 0);
```
arutk, line 397: I suggest to change the code so that return code from env_spinlock_init is propagated up:

line 392 should look like this:
result =env_spinlock_init(&freelist->lock[i])
if (result)
    goto spinlock_err;

arutk: line 388:
if (!freelist->lock || !freelist->part) {
   result = -OCF_ERR_NO_MEM;
   goto free_allocs;
}

src/ocf_freelist.c
```diff
@@ -406,7 +407,7 @@ ocf_freelist_t ocf_freelist_init(struct ocf_cache *cache)
 	env_vfree(freelist->lock);
 	env_vfree(freelist->part);
 	env_vfree(freelist);
-	return NULL;
+	return -OCF_ERR_NO_MEM;
```
arutk, line 410: reeturn result;

arutk (changes requested):

src/mngt/ocf_mngt_cache.c
```diff
@@ -1791,36 +1834,47 @@ static void _ocf_mngt_cache_attach(ocf_cache_t cache,
...
 
-	result = env_memcpy(data, cfg->uuid.size, cfg->uuid.data,
-			cfg->uuid.size);
-	if (result)
-		goto err_uuid;
+	OCF_PL_NEXT_RET(pipeline);
+}
 
-	context->cfg.uuid.data = data;
+static void _ocf_mngt_cache_load(ocf_cache_t cache,
```
robertbaldyga, line 1835: Why don't just move contents of this function to `ocf_mngt_cache_load()`? And the same about attach. Those functions are now really small and this additional indirection level brings no value, it only makes code more difficult to analyze.

And actually as we have separate load/attach paths, we could also get rid of `_ocf_mngt_cache_attach_complete()`/`_ocf_mngt_cache_load_complete()` and just move them to pipeline finish. The less callback/priv pairs the better for code readability.

src/mngt/ocf_mngt_cache.c
robertbaldyga: Also be careful with freeing `data` in pipeline finish, because if error occurred in line 1299 it will result in double free. Setting `context->cfg.uuid.data` in error path will be probably the simples solution.

robertbaldyga (changes requested):

src/ocf_freelist.c
```diff
@@ -365,7 +365,7 @@ void ocf_freelist_put_cache_line(ocf_freelist_t freelist,
 	env_put_execution_context(ctx);
 }
 
-ocf_freelist_t ocf_freelist_init(struct ocf_cache *cache)
+int ocf_freelist_init(ocf_freelist_t *flst, struct ocf_cache *cache)
```
robertbaldyga, line 368: By convention when we return by pointer we typically name local pointer something like `tmp_freelist` and the function argument pointer `freelist`.

robertbaldyga (changes requested):

src/mngt/ocf_mngt_cache.c
arutk: this step is also used in ocf_mngt_cache_attach, it is moved to pipeline for the purpose of reusing in both attach and load

arutk (approved):

robertbaldyga (approved):

PR #433, opened 2020-12-28
Parallel eviction

arutk:


src/eviction/lru.c
```diff
@@ -384,71 +547,79 @@ bool evp_lru_can_evict(ocf_cache_t cache)
...
+	bool cl_write_lock =
+		(req->engine_cbs->get_lock_type(req) ==	ocf_engine_lock_write);
+	unsigned evp;
+	unsigned req_idx = 0;
 
 	if (cline_no == 0)
 		return 0;
 
-	lru_iter_init(&iter, cache, part, part->next_eviction_list, true);
+	evp = env_atomic64_inc_return(&part->curr_evp) %
```
arutk, line 568: perf queue counter

src/utils/utils_part.c
```diff
@@ -191,81 +191,25 @@ void ocf_part_set_valid(struct ocf_cache *cache, ocf_part_id_t id,
...
-	_to_evict = ocf_part_evict_size(req);
-
-	if (_to_evict == 0)
-		ret = OCF_PART_HAS_SPACE;
-
-	if (to_evict)
-		*to_evict = _to_evict;
-
-	return ret;
+	return (part_occupancy + needed_cache_lines <= part_occupancy_limit);
```
arutk, line 215: potential integer overflow

src/concurrency/ocf_metadata_concurrency.c
```diff
@@ -140,36 +150,62 @@ void ocf_metadata_concurrency_attached_deinit(
...
+	return error;
 }
 
 void ocf_metadata_end_exclusive_access(
 		struct ocf_metadata_lock *metadata_lock)
 {
-        env_rwsem_up_write(&metadata_lock->global);
+	unsigned i;
+
+	for (i = OCF_NUM_GLOBAL_META_LOCKS; i > 0; i--)
```
arutk, line 186: iterate from zero up

src/eviction/lru.c
arutk: fixed

src/utils/utils_part.c
arutk: fixed in the other PR

src/eviction/lru.c
arutk: done

src/eviction/lru.c
```diff
@@ -238,140 +278,242 @@ static inline bool _lru_evp_all_empty(struct ocf_lru_iter *iter)
...
 
-	ocf_cleaner_fire(cache, &attribs);
-}
+	ocf_metadata_start_shared_access(&cache->metadata.lock);
 
-static void evp_lru_zero_line_complete(struct ocf_request *ocf_req, int error)
-{
-	env_atomic_dec(&ocf_req->cache->pending_eviction_clines);
-}
+	OCF_METADATA_EVICTION_LOCK_ALL();
```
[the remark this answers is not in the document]
arutk: there is no visible performance degrdation after reaching 100%

src/concurrency/ocf_cache_line_concurrency.h
```diff
@@ -153,6 +153,14 @@ void ocf_req_unlock_entry(struct ocf_cache *cache,
  */
 void ocf_cache_line_unlock_rd(struct ocf_cache *cache, ocf_cache_line_t line);
 
+/**
+ * @brief Release cache line write lock
+ *
+ * @param cache - OCF cache instance
+ * @param line - Cache line to be unlocked
+ */
+void ocf_cache_line_unlock_wr(struct ocf_cache *cache, ocf_cache_line_t line);
```
robertbaldyga, line 162: This is duplicate declaration (see line 181).

robertbaldyga (changes requested):
In commit message of "Add LOOKUP_REMAPPED status to allow iterative cacheline lock":
s/cacheline lock irequest called/cacheline lock called/

src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -715,6 +715,14 @@ static inline void __remove_line_from_waiters_list(struct ocf_cache_line_concurr
 	__unlock_waiters_list(c, line, flags);
 }
 
+static inline bool _ocf_req_needs_cl_lock(struct ocf_request *req, unsigned entry)
+{
+	/* remapped cachelines are assigned cacheline lock individually
+	   during eviction */
```
robertbaldyga, line 721: Incorrect comment formatting.

robertbaldyga (changes requested):
In commit message of "Rename cleaner attribute cache_line_lock to lock_cacheline":
s/make it clean/make it clear/

src/utils/utils_cleaner.h
```diff
@@ -22,11 +22,14 @@ typedef int (*ocf_cleaner_get_item)(struct ocf_cache *cache,
 		void *getter_context, uint32_t item,
 		ocf_cache_line_t *line);
 
+typedef void (*ocf_cleaner_put_item)(struct ocf_cache *cache,
+		void *getter_context, uint32_t item);
+
```
robertbaldyga: It doesn't look like it's meant to be part of this commit.

robertbaldyga (changes requested):

src/concurrency/ocf_cache_line_concurrency.c
arutk: fixed

src/utils/utils_cleaner.h
arutk: right

src/engine/engine_common.c
```diff
@@ -122,36 +147,38 @@ void ocf_engine_update_req_info(struct ocf_cache *cache,
...
+	if (status == LOOKUP_MISS)
+		return;
+
+	/* Check if cacheline is sequential with previous */
+	if (entry > 0 && ocf_engine_clines_phys_cont(req, entry - 1))
+		req->info.seq_no++;
+	/* Check if cacheline is sequential with next */
+	if (entry < req->core_line_count - 1 &&
+			ocf_engine_clines_phys_cont(req, entry)) {
+		req->info.seq_no++;
```
arutk, line 181: we need tests for this stuff

src/utils/utils_cleaner.c
```diff
@@ -333,8 +332,12 @@ static int _ocf_cleaner_update_metadata(struct ocf_request *req)
 
 		cache_line = iter->coll_idx;
 
+		ocf_hb_cline_prot_lock_wr(&cache->metadata.lock,
+				req->lock_idx, req->map[i].core_id,
+				req->map[i].core_line);
+
 		if (!metadata_test_dirty(cache, cache_line))
-			continue;
+			goto unlock;
```
imjfckm, line 341: Jumping inside the loop? Maybe invert the logic:
```c
if (metadata_test_dirty(cache, cache_line)) {
        <do_stuff>
}
<unlock>
```

src/engine/engine_common.c
```diff
@@ -93,18 +93,43 @@ static inline int _ocf_engine_check_map_entry(struct ocf_cache *cache,
...
+{
+	struct ocf_map_info *entry1, *entry2;
+	ocf_cache_line_t phys1, phys2;
+
+	entry1 = &req->map[entry];
+	entry2 = &req->map[entry + 1];
+
+	if (!entry1->status || entry1->status == LOOKUP_MISS ||
+			!entry2->status || entry2->status == LOOKUP_MISS)
+	{
```
imjfckm, line 110: All status codes are non-zero. I think in its current call site the code would sooner BUG-out than `!entry1->status` is `true`

imjfckm: also, `entry1->status` can't be true here, so entire check pertaining to entry1 is redundand

src/engine/engine_common.c
imjfckm, line 181: I don't get it - you're incrementing `seq_no` up to twice per map entry. This way if you have 3 sequential core lines in request you'd get `seq_no` equal to 4, which makes no sense. Is there something I'm missing?

src/engine/engine_common.c
arutk: entry->status may be zero in case of entries which weren't traversed yet

arutk: The idea of this change is to have correct seq_no value after request is traversed (sets all the statuses to HIT or MISS) and then assigned cachelines in eviction (REMAPPED). During the initial traversation entry2->status will be null.

src/engine/engine_common.c
arutk: not twice, during initial traversation entry2->status will always be 0, so each execution of this piece of code will increment the counter at most once.

arutk: However now I see that I didn't consider the fact, that the tracersation itself is called twice, so the counter might be incremented too many times.

arutk: Now I see a problem with my implementation - traversation is performed twice (initial and before eviction). map->status is not zeroed before each traversation. So there will be stale status in map[entry+1] during the second traversation.

robertbaldyga (changes requested):

src/eviction/lru.c
```diff
@@ -238,140 +278,243 @@ static inline bool _lru_evp_all_empty(struct ocf_lru_iter *iter)
...
+		while (cline != end_marker && !_lru_iter_evition_lock(iter,
+				cline, core_id, core_line)) {
+			cline = ocf_metadata_get_eviction_policy(
+					iter->cache, cline)->lru.prev;
 		}
 
-		ENV_BUG_ON(!metadata_test_dirty(cache, cline));
+		if (cline != end_marker) {
+			remove_lru_list(cache, list, cline);
+			add_lru_head(cache, list, cline);
```
robertbaldyga, line 381: Need to rebalance the list.

robertbaldyga (changes requested):

arutk:
retest this please

src/engine/engine_common.c
```diff
@@ -93,18 +93,43 @@ static inline int _ocf_engine_check_map_entry(struct ocf_cache *cache,
 		return -1;
 }
 
+/* Returns true if core lines no 'entry' and 'entry + 1' within the request
```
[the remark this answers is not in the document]
arutk: Actually I meant "no" like "number". I think we can agree on "core lines on index 'entry' and 'entry + 1'?

src/engine/engine_common.c
arutk: Ok, I am adjusting the code so that this "double check" (for previous and and next cacheline being sequential with current) is performed explicitly only for cachelines that are "patched" as a result of insertion via eviction. So the existing traversation, mapping and "update_req" functions would have minimal changes in sequential detection logic. 
And eviction code would need to additionally call "patch_req" to update stats for the remapped cacheline. Also there would be no checks for "entry->status != 0", as these were not needed in the baseline and now patch_req() is always called with request fully traversed, meaning all the cachelines have a proper tatus set.

src/engine/engine_common.c
arutk: removing all these checks

src/ocf_request.h
```diff
@@ -103,6 +103,8 @@ struct ocf_request {
 	struct ocf_io_internal ioi;
 	/*!< OCF IO associated with request */
 
+	const struct ocf_engine_callbacks *engine_cbs;
+
```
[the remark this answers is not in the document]
arutk: ok, I agree. but let's do the renaming as a separate PR. I am adding the comment per your request.

src/utils/utils_cleaner.c
arutk: ok :)

robertbaldyga (changes requested):

src/engine/engine_common.c
```diff
@@ -93,66 +93,100 @@ static inline int _ocf_engine_check_map_entry(struct ocf_cache *cache,
...
 			 * Need to move this cache line into other partition
 			 */
-			_entry->re_part = true;
+			entry->re_part = true;
 			req->info.re_part_no++;
 		}
 
 		break;
+	case LOOKUP_INSERTED:
+		req->info.insert_no++;
```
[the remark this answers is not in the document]
arutk: right, it wasn't intended

src/engine/engine_common.c
```diff
@@ -93,66 +93,100 @@ static inline int _ocf_engine_check_map_entry(struct ocf_cache *cache,
...
 	case LOOKUP_MISS:
-		req->info.seq_req = false;
 		break;
-	case LOOKUP_INSERTED:
 	case LOOKUP_REMAPPED:
-		break;
+		/* remapped cachelines are to be updated via
+		 * ocf_engine_patch_req_info()
+		 */
```
robertbaldyga, line 190: This fall through will trigger BUG_ON() on request resume.

robertbaldyga (dismissed):

src/engine/engine_common.c
```diff
@@ -521,12 +607,23 @@ int ocf_engine_prepare_clines(struct ocf_request *req,
 	promote = ocf_promotion_req_should_promote(
 			req->cache->promotion_policy, req);
 	if (!promote) {
+		ocf_engine_set_hot(req);
```
arutk, line 610: unnecessary - request is about to go in pt

src/engine/engine_common.c
```diff
@@ -410,94 +472,118 @@ static void _ocf_engine_clean_end(void *private_data, int error)
...
+/* Attempt to map cachelines marked as LOOKUP_MISS by evicting from cache.
+ * Caller must assure that request map info is up to date (request
+ * is traversed).
+ */
+static inline int ocf_prepare_clines_evict(struct ocf_request *req)
+{
+	int lock_status = -OCF_ERR_NO_LOCK;
+	bool part_has_space;
+
+	ocf_engine_set_hot(req);
```
arutk, line 521: move closer to function exit (level up) - merge all set_hot into 1

arutk: set_hot can be done outside hash bucket locks

src/eviction/lru.c
```diff
@@ -454,71 +598,78 @@ bool evp_lru_can_evict(ocf_cache_t cache)
...
+		/* TODO: if atomic mode is restored, need to zero metadata
+		 * before proceeding with cleaning (see version <= 20.12) */
+
+		/* find next unmapped cacheline in request */
+		while (req_idx + 1 < req->core_line_count &&
+				req->map[req_idx].status != LOOKUP_MISS) {
+			req_idx++;
 		}
 
+		ENV_BUG_ON(req->map[req_idx].status != LOOKUP_MISS);
```
mmichal10, line 644: If the last core line from request would have status different than `LOOKUP_MISS`, `BUG_ON()` will be triggered

src/utils/utils_cleaner.c
```diff
@@ -898,12 +899,23 @@ void ocf_cleaner_fire(struct ocf_cache *cache,
...
+				&core_sector);
+
+		if (attribs->lock_metadata) {
+			ocf_hb_cline_prot_lock_rd(&cache->metadata.lock,
+					req->lock_idx, core_id, core_sector);
+		}
+
+		skip = false;
+
+		/* when line already cleaned - rare condition under h
```
mmichal10, line 913: You lost `eavy` after `h`

src/utils/utils_cleaner.c
```diff
@@ -914,12 +926,16 @@ void ocf_cleaner_fire(struct ocf_cache *cache,
 			 * Cache line (sector) cannot be dirty and not valid
 			 */
 			ENV_BUG();
-			continue;
+			skip = true;
```
mmichal10, line 929: I believe setting `skip`  can be skipped because of `BUG()` in previous line

src/eviction/lru.c
arutk: that's right. this would mean that the caller requested to evict more cachelines than there are unmapped cachelines within the request.

mmichal10 (dismissed):

mmichal10 (approved):

imjfckm (approved):

mmichal10 (approved):

PR #445, opened 2021-01-28
probe: return dirty and shutdown status despite metadata mismatch

arutk:


Ostrokrzew (dismissed):

imjfckm (approved):

src/mngt/ocf_mngt_cache.c
```diff
@@ -2050,7 +2064,7 @@ static void _ocf_mngt_cache_unplug(ocf_cache_t cache, bool stop,
 	__deinit_cleaning_policy(cache);
 	__deinit_promotion_policy(cache);
 
-	if (ocf_mngt_cache_is_dirty(cache)) {
+	if (dirty) {
```
robertbaldyga, line 2067: Why can't we just do this earlier (before removing cores)?

arutk: we can, but it seems to belong here logically - don't you think?

robertbaldyga: Imo we could even do it as a first thing during cache stop. We don't expect any flushing at that point anyway.

robertbaldyga (approved):

jfckm (approved):

arutk:
retest this please

PR #446, opened 2021-01-29
Don't evict on hit

mmichal10:
If request is hit, simply try to acquire cachelines instead of verifying
whether target partition's size is not exceeded.


mmichal10:
Fixes https://github.com/Open-CAS/open-cas-linux/issues/663

arutk (approved):

robertbaldyga (approved):

PR #452, opened 2021-02-04
Split global metadata lock

arutk:
Divide single global lock instance into 4 to reduce contention in multiple read-locks scenario.

### DIFF
--- a/src/concurrency/ocf_metadata_concurrency.c
+++ b/src/concurrency/ocf_metadata_concurrency.c
@@ -155,14 +155,12 @@ void ocf_metadata_end_exclusive_access(
         env_rwsem_up_write(&metadata_lock->global);
 }
 
-void ocf_metadata_start_shared_access(
-		struct ocf_metadata_lock *metadata_lock)
+void ocf_metadata_start_shared_access(struct ocf_metadata_lock *metadata_lock)
 {
         env_rwsem_down_read(&metadata_lock->global);
 }
 
-int ocf_metadata_try_start_shared_access(
-		struct ocf_metadata_lock *metadata_lock)
+int ocf_metadata_try_start_shared_access(struct ocf_metadata_lock *metadata_lock)
 {
 	return env_rwsem_down_read_trylock(&metadata_lock->global);
 }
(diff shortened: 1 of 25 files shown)


src/concurrency/ocf_metadata_concurrency.c
```diff
@@ -215,10 +215,8 @@ int ocf_metadata_hash_try_lock(struct ocf_metadata_lock *metadata_lock,
 		ENV_BUG();
 	}
 
-	if (!result)
-		return -1;
 
-	return 0;
+	return result;
```
robertbaldyga, line 282: This changes meaning of the result without update in places when this function is used. In result this commit breaks bisectability.

robertbaldyga (changes requested):

src/concurrency/ocf_metadata_concurrency.c
arutk: actually this function is not referenced anywhere at this point. it starts being used in the next commit, also being changed to "status"

imjfckm (dismissed):

src/concurrency/ocf_metadata_concurrency.c
```diff
@@ -140,34 +151,69 @@ void ocf_metadata_concurrency_attached_deinit(
...
+   picking wisely would allow for higher read througput:
+   * free running per-cpu counter sounds good,
+   * for rarely excercised code paths (e.g. management) any value would do.
+*/
+void ocf_metadata_start_shared_access(
+		struct ocf_metadata_lock *metadata_lock,
+		unsigned lock_idx)
 {
-        env_rwsem_down_read(&metadata_lock->global);
+        env_rwsem_down_read(&metadata_lock->global[lock_idx % OCF_NUM_GLOBAL_META_LOCKS]);
```
robertbaldyga, line 202: Why doesn't `ocf_queue_get_next_shard_idx()` ensure that `lock_idx` is within range?

robertbaldyga (changes requested):

src/mngt/ocf_mngt_cache.c
```diff
@@ -2344,11 +2344,11 @@ ocf_promotion_t ocf_mngt_cache_promotion_get_policy(ocf_cache_t cache)
 {
 	ocf_promotion_t result;
 
-	ocf_metadata_start_shared_access(&cache->metadata.lock);
+	ocf_metadata_start_shared_access(&cache->metadata.lock, 0);
```
robertbaldyga, line 2347: `lock_idx = ocf_queue_get_next_shard_idx(cache-mngt_queue);` ?

robertbaldyga (changes requested):

src/concurrency/ocf_metadata_concurrency.c
arutk: done

src/mngt/ocf_mngt_cache.c
arutk: I think this is so rare that it isn't worth the effort

robertbaldyga (dismissed):

imjfckm (approved):

imjfckm (approved):

robertbaldyga (approved):

PR #453, opened 2021-02-05
Skip cacheline concurrency global lock in fast path

arutk:
The main purpose of cacheline concurrency global lock
is to eliminate the possibility of deadlocks when
locking multiple cachelines.

Cacheline lock fast path does not need to acquire
this lock, as it is only opportunistically attempting
to lock all clines without wait. There is no risk
of deadlock, as:
 * concurrent fast path will also only try_lock
   cachelines, releasing all acquired locks if failed
   to immediately acquire lock for any cacheline
 * concurrent slow path is guaranteed to have
   precedence in lock acquisition when conditions
   for deadlock occure (both slowpath and fastpath
   have acquired some locks required by the other
   thread). This is because the fastpath thread will
   back off (release acquired locks) if any one of the
   cacheline locks is not acquired.

### DIFF
--- a/src/concurrency/ocf_cache_line_concurrency.c
+++ b/src/concurrency/ocf_cache_line_concurrency.c
@@ -50,10 +50,12 @@ struct __waiters_list {
 };
 
 struct ocf_cache_line_concurrency {
+	ocf_cache_t cache;
 	env_mutex lock;
 	env_atomic *access;
 	env_atomic waiting;
 	size_t access_limit;
+	ocf_cache_line_t num_clines;
 	env_allocator *allocator;
 	struct __waiters_list waiters_lsts[_WAITERS_LIST_ENTRIES];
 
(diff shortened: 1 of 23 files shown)


imjfckm (dismissed):

src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -833,9 +833,7 @@ int ocf_req_async_lock_rd(struct ocf_request *req, ocf_req_async_lock_cb cb)
 		req->cache->device->concurrency.cache_line;
 	int lock;
 
-	env_rwsem_down_read(&c->lock);
 	lock = _ocf_req_trylock_rd(req);
```
arutk, line 836: After consulting with Mariusz: looks like there is a potential for deadlock in this change when fastpath wins race with concurrent slowpath and then releases the problematic lock in fastpath rollback without transfering lock ownership to the first waiter.

arutk:
retest this please

src/concurrency/ocf_cache_line_concurrency.c
arutk: if's fixed now + tests are added to validate

mariuszbarczak (approved):

rafalste (approved):

mariuszbarczak (approved):

rafalste (approved):

imjfckm (commented):
🚀

imjfckm (approved):

PR #457, opened 2021-02-08
Fix broken 'dirty_for' stats

Ostrokrzew:
Connected with https://github.com/Open-CAS/open-cas-linux/issues/680

### DIFF
--- a/src/metadata/metadata.c
+++ b/src/metadata/metadata.c
@@ -1192,8 +1192,9 @@ static void _recovery_rebuild_cline_metadata(ocf_cache_t cache,
 		env_atomic_inc(&core->runtime_meta->dirty_clines);
 		env_atomic_inc(&core->runtime_meta->
 				part_counters[part_id].dirty_clines);
-		env_atomic64_cmpxchg(&core->runtime_meta->dirty_since,
-				0, env_get_tick_count());
+		if (!env_atomic64_read(&core->runtime_meta->dirty_since))
+			env_atomic64_cmpxchg(&core->runtime_meta->dirty_since, 0,
+					env_ticks_to_secs(env_get_tick_count()));
 	}
 }
 
--- a/src/ocf_cache.c
+++ b/src/ocf_cache.c
@@ -56,11 +56,11 @@ ocf_cache_mode_t ocf_cache_get_mode(ocf_cache_t cache)
 	return cache->conf_meta->cache_mode;
 }
 
-static uint32_t _calc_dirty_for(uint64_t dirty_since)
+static uint64_t _calc_dirty_for(uint64_t dirty_since)
 {
-	return dirty_since ?
-		(env_ticks_to_msecs(env_get_tick_count() - dirty_since) / 1000)
-		: 0;
+	uint64_t current_time = env_ticks_to_secs(env_get_tick_count());
+
+	return dirty_since ? (current_time - dirty_since) : 0;
 }
 
 int ocf_cache_get_info(ocf_cache_t cache, struct ocf_cache_info *info)
--- a/src/ocf_core.c
+++ b/src/ocf_core.c
@@ -142,11 +142,11 @@ int ocf_core_visit(ocf_cache_t cache, ocf_core_visitor_t visitor, void *cntx,
 
 /* *** HELPER FUNCTIONS *** */
 
-static uint32_t _calc_dirty_for(uint64_t dirty_since)
+static uint64_t _calc_dirty_for(uint64_t dirty_since)
 {
-	return dirty_since ?
-		(env_ticks_to_msecs(env_get_tick_count() - dirty_since) / 1000)
-		: 0;
+	uint64_t current_time = env_ticks_to_secs(env_get_tick_count());
+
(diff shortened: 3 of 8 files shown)


src/ocf_cache.c
```diff
@@ -56,11 +56,11 @@ ocf_cache_mode_t ocf_cache_get_mode(ocf_cache_t cache)
...
 
-static uint32_t _calc_dirty_for(uint64_t dirty_since)
+static uint64_t _calc_dirty_for(uint64_t dirty_since)
 {
-	return dirty_since ?
-		(env_ticks_to_msecs(env_get_tick_count() - dirty_since) / 1000)
-		: 0;
+	uint64_t current_time = env_ticks_to_secs(env_get_tick_count());
+
+	return dirty_since ? ((current_time - dirty_since) % ULONG_MAX) : 0;
```
arutk, line 63: I think "% ULONG_MAX" is not needed here (applies to _calc_dirty_for as well)

src/utils/utils_cache_line.c
```diff
@@ -157,7 +157,7 @@ void set_cache_line_dirty(struct ocf_cache *cache, uint8_t start_bit,
 			 * If this is first dirty cline set dirty timestamp
 			 */
 			env_atomic64_cmpxchg(&req->core->runtime_meta->dirty_since,
```
arutk, line 159: env_get_tick_count (acquiring monotonic high precission timestamp) is potentially slow. it would make sense to first read "dirty_since" and attempt cmpxchg only if it was 0

arutk: I think I would rename this function to "env_get_time_ns", since this has quite different semantics than ticks.

src/ocf_core.c
```diff
@@ -142,11 +142,11 @@ int ocf_core_visit(ocf_cache_t cache, ocf_core_visitor_t visitor, void *cntx,
...
 
-static uint32_t _calc_dirty_for(uint64_t dirty_since)
+static uint64_t _calc_dirty_for(uint64_t dirty_since)
 {
-	return dirty_since ?
-		(env_ticks_to_msecs(env_get_tick_count() - dirty_since) / 1000)
-		: 0;
+	uint64_t current_time = env_ticks_to_secs(env_get_tick_count());
+
+	return dirty_since ? ((current_time - dirty_since) % ULONG_MAX) : 0;
```
arutk, line 149: how does this compare to env_time_after?

arutk (changes requested):

src/ocf_core.c
Ostrokrzew: `env_time_after` takes 2 agruments and returns boolean value - is parameter A value greater than parameter B value? Not applicable here.

src/utils/utils_cache_line.c
Ostrokrzew: AD 1. I did the check before exchanging value.
AD 2. I think the names in this group of functions (time-related) are the same as in our POSIX-compatible `ocf_env` header. Of course we can discuss it.

robertbaldyga (changes requested):
Commits "Reformat function that calculates how long" and "Replace seconds with seconds" should be merged into one, as the first one breaks the code assuming, that `dirty_for` is already in seconds.

src/metadata/metadata.c
```diff
@@ -1192,8 +1192,9 @@ static void _recovery_rebuild_cline_metadata(ocf_cache_t cache,
 		env_atomic_inc(&core->runtime_meta->dirty_clines);
 		env_atomic_inc(&core->runtime_meta->
 				part_counters[part_id].dirty_clines);
-		env_atomic64_cmpxchg(&core->runtime_meta->dirty_since,
-				0, env_get_tick_count());
+		if (!env_atomic64_read(&core->runtime_meta->dirty_since))
+			env_atomic64_xchg(&core->runtime_meta->dirty_since,
```
arutk, line 1196: I think the previous version (env_atomic64_cmpxchg) was more precise

src/ocf_core.c
arutk: ok, thanks

src/metadata/metadata.c
Ostrokrzew: Then double check would be done - first we check it and then `atomic_cmpxch` would do check. Wouldn't it be redundant?

arutk: I think it would be ok.  First of all the double check would be relatively rare, it happens only when dirty counter is incremented from zero. Secondly the proper way of not double-checking would be to do it the old way (only env_atomic64_cmpxchg without any condition) - but this might have a real perf impact every I/O. I believe the double check that I suggest would be a good compromise between performance and precision (there would be no time-of-check vs time of use mismatch), although the difference between this and the current version is probably negligible. You know my opinion, please use your judgement :).

Ostrokrzew: Then the new wrapper in the CAS for the atomic function would be unnecessary and IMHO less new code is better. Thanks for that explanation.

arutk (approved):

rafalste (approved):

robertbaldyga (approved):

PR #458, opened 2021-02-09
Fix updating hot cachelines cleaning list

### LINKED ISSUE #455: Cache lines' access timestamp in cleaning policy metadata not being updated
`ocf_cleaning_set_hot_cache_line()` should be called each time a dirty cache line is accessed. Instead it is called only when the cache line is changing it's state from clean to dirty.

### How to reproduce:
- Start a cache in WB and configure alru `casadm -X -n cleaning-alru -i 1 -w 0 -s 30 -b 100 -t 0`
- Issue writes to single cache line `fio --name=nvme --numjobs=1 --iodepth=4 --bs=4k --rw=write --ioengine=libaio --direct=1 --group_reporting=1 --filename=/dev/cas1-1 --filesize=4k --runtime=60 --time_based=1`
- Dirty block flushes as soon as fio is completed 

### Expected result:
Dirty block flushed 30 seconds (`--staleness-time 30`) after fio is completed.

mmichal10:
Update cacheline's timestamp each time it's being written.

Fixes #455 

### DIFF
--- a/src/engine/engine_wb.c
+++ b/src/engine/engine_wb.c
@@ -13,6 +13,7 @@
 #include "../ocf_request.h"
 #include "../utils/utils_io.h"
 #include "../utils/utils_cache_line.h"
+#include "../utils/utils_request.h"
 #include "../utils/utils_part.h"
 #include "../concurrency/ocf_concurrency.h"
 
@@ -42,6 +43,8 @@ static void _ocf_write_wb_update_bits(struct ocf_request *req)
 
 		ocf_req_hash_unlock_wr(req);
 	}
+
+	ocf_req_set_cleaning_hot(req);
 }
 
 static void _ocf_write_wb_io_flush_metadata(struct ocf_request *req, int error)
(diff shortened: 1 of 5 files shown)


src/engine/engine_wb.c
```diff
@@ -42,6 +43,9 @@ static void _ocf_write_wb_update_bits(struct ocf_request *req)
 
 		ocf_req_hash_unlock_wr(req);
 	}
+
+	if (ocf_engine_mapped_count(req) > 0)
```
arutk, line 47: So you will not call "set_hot" in case of full miss? I think it's incorrect.

arutk: Ok, it actually might make sense to keep track of write operations for the purpose of cleaning

arutk (approved):

src/engine/engine_wb.c
mmichal10: Ok, I've changed the code to call set_hot() unconditionally because I couldn't came up with any scenario when a write IO would be submitted in WB and the timestamp shouldn't be updated.

arutk (approved):

imjfckm (approved):

PR #459, opened 2021-02-11
Add error according to new core management operation

Ostrokrzew:
Add error to pyocf test utilities

Signed-off-by: Slawomir Jankowski <slawomir.jankowski@intel.com>

Required for https://github.com/Open-CAS/open-cas-linux/pull/690

### DIFF
--- a/inc/ocf_err.h
+++ b/inc/ocf_err.h
@@ -125,6 +125,9 @@ typedef enum {
 
 	/** Device does not meet requirements */
 	OCF_ERR_INVAL_CACHE_DEV,
+
+	/** Core device is in active state */
+	OCF_ERR_CORE_IN_ACTIVE_STATE,
 } ocf_error_t;
 
 #endif /* __OCF_ERR_H__ */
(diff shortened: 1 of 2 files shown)


inc/ocf_err.h
```diff
@@ -125,6 +125,9 @@ typedef enum {
 
 	/** Device does not meet requirements */
 	OCF_ERR_INVAL_CACHE_DEV,
+
+	/** Core device is in active state */
+	OCF_ERR_CORE_IN_ACTIVE_STATE,
```
arutk, line 130: Is it going to be used anywhere? If so, it would be good to see both this commit and the one with actual usage in one PR so that we have a full picture

Ostrokrzew: Will be used in OCL. See PR mentioned in description.

arutk: sorry, I missed that. but in this case this should be defined in OCL, there is a separate set of OCL error codes (i think "KCAS_ERR_...."

PR #462, opened 2021-02-16
Get rid of status bits lock

arutk:
There is no purpose in maintaining global R/W lock for status bits while we can protect the relevant pieces of metadata with hash bucket locks.


rafalste (approved):

Ostrokrzew (approved):
Very nice cleaning and clear info in commits. I like it!
You've got typo in second commit message.

mmichal10 (approved):

robertbaldyga (approved):

PR #465, opened 2021-02-26
Use mpool allocators for requests and metadata_io

jfckm:
metadata_io previously used single, statically sized allocation for all its IO needs. This was sub-optimal not only because of the overallocation, but also because it was allocated using `vmalloc`. This patch changes that to mpool allocators which enables OCL to back those allocations with slab allocations and CAS's own reserve pools.
To avoid huge memory footprint increase from this change I also limited rpool size for higher-order allocations in metadata_io mpools.

This change is dependent on https://github.com/Open-CAS/open-cas-linux/pull/712.

### DIFF
--- a/src/metadata/metadata_io.c
+++ b/src/metadata/metadata_io.c
@@ -179,10 +179,11 @@ int metadata_io_read_i_atomic(ocf_cache_t cache, ocf_queue_t queue, void *priv,
 static void metadata_io_req_fill(struct metadata_io_request *m_req)
 {
 	ocf_cache_t cache = m_req->cache;
+	struct metadata_io_request_asynch *a_req = m_req->asynch;
 	int i;
 
 	for (i = 0; i < m_req->count; i++) {
-		m_req->on_meta_fill(cache, m_req->data,
+		a_req->on_meta_fill(cache, m_req->data,
 			m_req->page + i, m_req->context);
 	}
 }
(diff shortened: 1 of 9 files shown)


src/metadata/metadata_io.c
```diff
@@ -45,6 +45,19 @@ struct metadata_io_read_i_atomic_context {
...
+enum ocf_mio_size {
+	ocf_mio_size_1 = 0,
+	ocf_mio_size_2,
+	ocf_mio_size_4,
+	ocf_mio_size_8,
+	ocf_mio_size_16,
+	ocf_mio_size_32,
+	ocf_mio_size_64,
+	ocf_mio_size_max,
+};
```
arutk, line 57: Most of the enumeration values are unused. single const variable ocf_mio_size_max would be sufficient

arutk (dismissed):

src/metadata/metadata_io.c
```diff
@@ -455,6 +475,36 @@ int metadata_io_read_i_asynch(ocf_cache_t cache, ocf_queue_t queue,
...
 
+#define MIO_RPOOL_LIMIT 16
+#define MIO_RPOOL_THRESHOLD ocf_mio_size_16 /* This is statically determined to
+					       not exceed one page (4096B).
+					       Change if apropriate. */
+
+int ocf_metadata_io_ctx_init(struct ocf_ctx *ocf_ctx)
+{
+	uint32_t limits[] = {
+		[0 ... MIO_RPOOL_THRESHOLD] = -1,
```
arutk, line 486: [0 ... MIO_RPOOL_THRESHOLD - 1] ?

arutk (dismissed):

mmichal10 (approved):

arutk (approved):

PR #478, opened 2021-03-25
Fix freeing oversized discard requests

jfckm:
When issuing discard request over 512KiB OCF would trim this request and
overwrite req->core_line_count which would then cause this request to be
freed from wrong mpool.

This is fixed now by saving core_line_count that was set when allocating
this request that is never overwritten. This alloc_core_line_count is
then used to free the request from correct mpool.


rafalste:
ok to test

rafalste (approved):

arutk (approved):

mmichal10 (approved):

PR #490, opened 2021-04-14
Prevent adding core with the same UUID twice

mmichal10:

### DIFF
--- a/src/mngt/ocf_mngt_core.c
+++ b/src/mngt/ocf_mngt_core.c
@@ -27,6 +27,25 @@ static int ocf_mngt_core_set_name(ocf_core_t core, const char *name)
 			name, OCF_CORE_NAME_SIZE);
 }
 
+static int ocf_core_get_by_uuid(ocf_cache_t cache, void *uuid, size_t uuid_size,
+		ocf_core_t *core)
+{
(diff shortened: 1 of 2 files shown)


rafalste (approved):

rafalste (approved):

src/mngt/ocf_mngt_core.c
```diff
@@ -27,6 +27,25 @@ static int ocf_mngt_core_set_name(ocf_core_t core, const char *name)
...
 
+static int ocf_core_get_by_uuid(ocf_cache_t cache, void *uuid, size_t uuid_size,
+		ocf_core_t *core)
+{
+	struct ocf_volume *volume;
+	ocf_core_t i_core;
+	ocf_core_id_t i_core_id;
+
+	for_each_core_metadata(cache, i_core, i_core_id) {
+		volume = ocf_core_get_volume(i_core);
```
arutk, line 38: is it safe to get volume for a core which is potentially detached? this would probably happen during startup

mmichal10: I had checked it with the following scenario and it worked:
image

arutk (approved):

PR #507, opened 2021-05-25
remove partition list

arutk:


mmichal10 (approved):

src/metadata/metadata.c
```diff
@@ -572,6 +571,7 @@ static int ocf_metadata_init_fixed_size(struct ocf_cache *cache,
 		cache->user_parts[i].runtime = &part_runtime[i];
 		cache->user_parts[i].id = i;
 	}
+	cache->free = &part_runtime[PARTITION_INVALID];
```
robertbaldyga, line 574: This looks terrible. We should have more meaningful define for that. And at one go get rid of this "+2" few lines above.

src/eviction/lru.c
```diff
@@ -224,49 +224,88 @@ static void balance_lru_list(ocf_cache_t cache,
...
+		struct ocf_part_runtime *dst_part, struct ocf_lru_list *dst_list,
+		ocf_part_id_t dst_part_id)
+{
+	remove_lru_list(cache, src_list, cline);
+	balance_lru_list(cache, src_list);
+	add_lru_head(cache, dst_list, cline);
+	balance_lru_list(cache, dst_list);
+	env_atomic_dec(&src_part->curr_size);
+	env_atomic_inc(&dst_part->curr_size);
+	ocf_metadata_set_partition_id(cache, cline, dst_part_id);
```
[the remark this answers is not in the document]
robertbaldyga: That's a good point. Even though balancing is pretty cheap, the freelist is used really often (and this is for both - eviction and mapping), so it may be worth to skip balancing for freelist. This btw can be done easily by adding `is_balanced` flag to part and setting it to `false` on freelist, and then just early returning from `balance_lru_list()` if the flag is set.

src/eviction/eviction.c
```diff
@@ -124,24 +125,32 @@ static inline uint32_t ocf_evict_partitions(ocf_cache_t cache,
...
 	ocf_part_id_t target_part_id = req->part_id;
 	struct ocf_user_part *target_part = &cache->user_parts[target_part_id];
 	uint32_t evict_cline_no = ocf_engine_unmapped_count(req);
-	uint32_t evicted;
+	uint32_t evicted = 0;
+
+	/* First attempt to map from freelist */
+	if (ocf_lru_num_free(cache) > 0) {
+		evicted = ocf_request_space(cache, req, cache->free,
+				PARTITION_INVALID, evict_cline_no);
```
robertbaldyga, line 139: The `free` partition is anything but invalid. We need better name.

src/eviction/lru.c
```diff
@@ -224,49 +224,88 @@ static void balance_lru_list(ocf_cache_t cache,
...
 {
-	ocf_part_id_t part_id = ocf_metadata_get_partition_id(cache, cline);
-	struct ocf_user_part *part = &cache->user_parts[part_id];
 	uint32_t ev_list = (cline % OCF_NUM_EVICTION_LISTS);
+	ocf_part_id_t part_id;
+	struct ocf_part_runtime *part;
+
+	part_id = ocf_metadata_get_partition_id(cache, cline);
+	ENV_BUG_ON(part_id == PARTITION_UNSPECIFIED ||
+			part_id == PARTITION_INVALID);
```
robertbaldyga, line 243: `ENV_BUG_ON(part_id > OCF_IO_CLASS_ID_MAX);` would do the trick.

src/eviction/lru.c
```diff
@@ -224,49 +224,88 @@ static void balance_lru_list(ocf_cache_t cache,
...
+	ocf_part_id_t part_id = ocf_metadata_get_partition_id(cache, cline);
+	struct ocf_part_runtime *part = cache->user_parts[part_id].runtime;
 
 	list = evp_get_cline_list(cache, cline);
-	remove_lru_list(cache, list, cline);
-	balance_lru_list(cache, list);
+	free = evp_lru_get_list(cache->free, ev_list, true);
+
+	evp_lru_move(cache, cline, part, list, cache->free, free,
+			PARTITION_INVALID);
```
robertbaldyga, line 277: Again. Call for better name.

src/eviction/lru.c
```diff
@@ -636,14 +731,19 @@ uint32_t evp_lru_req_clines(struct ocf_request *req,
...
-	lru_iter_eviction_init(&iter, cache, part, evp, cl_write_lock, req);
+	lru_iter_eviction_init(&iter, cache, part, evp, req);
 
 	i = 0;
 	while (i < cline_no) {
 		if (!evp_lru_can_evict(cache))
 			break;
 
-		cline = lru_iter_eviction_next(&iter, &core_id, &core_line);
+		if (part_id != PARTITION_INVALID) {
```
robertbaldyga, line 741: Here too. TBH I didn't find a single use if `PARTITION_INVALID` that would really mean "invalid".

src/eviction/eviction.c
```diff
@@ -60,7 +59,8 @@ static inline uint32_t ocf_evict_part_do(struct ocf_request *req,
 		return 0;
 	}
 
-	return ocf_eviction_need_space(req->cache, req, target_part, to_evict);
+	return ocf_request_space(req->cache, req, target_part->runtime,
+		target_part->id, to_evict);
```
robertbaldyga, line 63: Maybe we can just have an `id` field inside the `runtime` to avoid passing this additional argument everywhere. Seems that it's strongly associated anyway.

src/eviction/ops.h
```diff
@@ -16,16 +16,16 @@
...
-	type = cache->conf_meta->eviction_policy_type;
+       type = cache->conf_meta->eviction_policy_type;
 
-	ENV_BUG_ON(type >= ocf_eviction_max);
+       ENV_BUG_ON(type >= ocf_eviction_max);
 
-	if (likely(evict_policy_ops[type].init_cline))
-		evict_policy_ops[type].init_cline(cache, line);
+       if (likely(evict_policy_ops[type].init_cline))
+               evict_policy_ops[type].init_cline(cache, line);
```
robertbaldyga, line 28: Please make it tabs again.

robertbaldyga (changes requested):

ArkadiuszNeumann:
retest this please

src/metadata/metadata.c
arutk: I would appreciate more constructive comments. INVALID makes some sense since the cachelines in this partition are invalid. However after some more thinking I realized that inserted cachelines can also be invalid, so FREE would probably be more precise.

src/eviction/ops.h
```diff
@@ -84,29 +84,28 @@ static inline void ocf_eviction_set_hot_cache_line(
...
 	}
 }
 
-static inline void ocf_eviction_flush_dirty(ocf_cache_t cache,
-		struct ocf_user_part *part, ocf_queue_t io_queue,
-		uint32_t count)
+static inline void ocf_lru_flush_dirty(ocf_cache_t cache,
+		struct ocf_part_runtime *part,
+		struct ocf_part_cleaning_ctx *ctx,
+		ocf_queue_t io_queue, uint32_t count)
```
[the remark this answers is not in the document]
arutk: let's do proper removal of evp abstraction and refactoring in a separate PR.

src/eviction/lru.c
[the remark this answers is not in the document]
arutk: done

src/eviction/ops.h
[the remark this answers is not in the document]
arutk: I have added removal of LRU list to this PR. However I think that ocf_lru_clean() belongs to this file as it is tightly based on LRU list.

src/eviction/eviction.c
arutk: added ocf_part structure to store runtime pointer as well as id for both user partitions and freelist

src/eviction/lru.c
arutk: after adding ocf_part struct this is changed to "part_id == PARTITION_UNSPECIFIED" as PARTITION_FREELIST is now a valid input to this function

robertbaldyga (changes requested):
- Typo in commit title "Separate ocf_part araray"
- Missing squash for commit "fix"

src/metadata/metadata.c
```diff
@@ -576,9 +576,9 @@ static int ocf_metadata_init_fixed_size(struct ocf_cache *cache,
 	for (i = 0; i < OCF_USER_IO_CLASS_MAX + 1; i++) {
 		cache->user_parts[i].config = &part_config[i];
 		cache->user_parts[i].clean_pol = &part_runtime_meta[i].clean_pol;
-		cache->user_parts[i].part.runtime =
-			&part_runtime_meta[i].runtime;
-		cache->user_parts[i].part.id = i;
+		cache->user_parts[i].part = &cache->parts[i];
```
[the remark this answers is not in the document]
arutk: moving ocf_part back to ocf_user_part

src/ocf_space.h
```diff
@@ -0,0 +1,41 @@
...
+/*
+ * Deallocates space according to eviction priorities.
+ *
+ * @returns:
+ * 'LOOKUP_HIT' if evicted enough cachelines to serve @req
+ * 'LOOKUP_MISS' otherwise
+ */
+int ocf_space_managment_remap_do(struct ocf_request *req);
+
+int space_management_free(ocf_cache_t cache, uint32_t count);
```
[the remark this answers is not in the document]
arutk: looks like this function is not defined at all - removing

robertbaldyga (approved):

arutk:
retest this please

mmichal10 (approved):

PR #509, opened 2021-06-01
Remove metadata updater

mmkayPL:


src/utils/utils_alock.c
```diff
@@ -0,0 +1,803 @@
...
+
+#define OCF_CACHE_LINE_ACCESS_WR	INT_MAX
+#define OCF_CACHE_LINE_ACCESS_IDLE	0
+#define OCF_CACHE_LINE_ACCESS_ONE_RD	1
+
+#define _WAITERS_LIST_SIZE	(16UL * MiB)
+#define _WAITERS_LIST_ENTRIES \
+	(_WAITERS_LIST_SIZE / sizeof(struct ocf_alock_waiters_list))
+
+#define _WAITERS_LIST_ITEM(cache_line) ((cache_line) % _WAITERS_LIST_ENTRIES)
```
arutk, line 42: "entry' would be better than "cache_line" here

arutk (approved):

arutk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/open-cas/ocf/509)* status: 0 of 2 LGTMs obtained

___
*[tests/unit/tests/concurrency/ocf_cache_line_concurrency.c/ocf_cache_line_concurrency.c, line 372 at r2](https://reviewable.io/reviews/open-cas/ocf/509#-MbBBe2p0n07iyhUE7uE:-MbBBe2q2WEVL3hMclSA:bgby002) ([raw file](https://github.com/open-cas/ocf/blob/b5d4ab07ea4b6b4579556b824bbc9b50c69aa3a4/tests/unit/tests/concurrency/ocf_cache_line_concurrency.c/ocf_cache_line_concurrency.c#L372)):*
> ```C
> 	{
> 		rw = rand() % 2;
> 		single = (rand() % 4 == 0);
> ```

let's make it %5 rather than %4 to make generate wider range of rw/single combinations.

src/utils/utils_alock.c
```diff
@@ -0,0 +1,803 @@
...
+		else
+			ocf_alock_unlock_one_wr_common(alock, entry);
+		ocf_alock_mark_index_locked(alock, req, i, false);
+	} else {
+		list_for_each_safe(iter, next, &lst->head) {
+			waiter = list_entry(iter, struct ocf_alock_waiter, item);
+			if (waiter->req == req) {
+				list_del(iter);
+				env_allocator_del(alock->allocator, waiter);
+			}
```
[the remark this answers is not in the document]
arutk: I think Michał is right, but there are some extra considerations - please double check them before adding a break here. One request may lock multiple cachelines resolving to the same 'idx' value. So there might be multiple waiter entries for a single request on a sinble list. The way we currently use this function assures that it is called once for each locked/waiting entry, so breaking early will yield some work to the next call. 

So I guess it boils down to how we define this function API.

src/ocf_request.h
```diff
@@ -222,9 +222,17 @@ struct ocf_request {
...
+
+	uint8_t *alock_entry_map;
+	/*!< Mapping for locked/unlocked alock entries */
+
 	struct ocf_map_info *map;
 
-	struct ocf_map_info __map[];
+	struct ocf_map_info __map[0];
+
+	uint8_t __alock_entry_map[0];
```
arutk, line 235: I think this isn't/shouldn't be used anywhere

arutk (dismissed):

KamilLepek:
retest this please

ArkadiuszNeumann:
retest this please

ArkadiuszNeumann:
retest this please

ArkadiuszNeumann:
retest this please

ArkadiuszNeumann:
retest this please

ArkadiuszNeumann:
retest this please

ArkadiuszNeumann:
retest this please

ArkadiuszNeumann:
retest this please

src/utils/utils_alock.h
```diff
@@ -20,29 +20,16 @@ struct ocf_alock;
...
 
 struct ocf_alock_lock_cbs
 {
-	ocf_cl_lock_line_needs_lock_cb line_needs_lock;
-	ocf_cl_lock_line_is_acting_cb line_is_acting;
-	ocf_cl_lock_line_is_locked_cb line_is_locked;
-	ocf_cl_lock_line_mark_locked_cb line_mark_locked;
-	ocf_cl_lock_line_get_entry_cb line_get_entry;
+	ocf_cl_lock_fast lock_lines_fast;
+	ocf_cl_lock_slow lock_lines_slow;
```
robertbaldyga, line 32: I think that `lock_entries_*` would be better, as the locked entries are not always cache lines. This also applies to naming convention in entire mio concurrency.

robertbaldyga (changes requested):

robertbaldyga (approved):

rafalste (approved):

PR #515, opened 2021-06-10
Use new non-zeroing allocator APIs

jfckm:


robertbaldyga (approved):

Ostrokrzew (approved):
Is it safe to remove allocator name validation?

PR #518, opened 2021-06-17
Reorder metadata updating patter in WB mode

### LINKED ISSUE #519: Improper status bits flushed to metadata
## Description
When an IO is issued to a cache which is in WB mode, cache line is being marked as valid and dirty before the data request is send to the disk. If different IO was handled in different thread it may flush metadata of the first IO before it sent it's data. If power shutdown appears in this very moment user will experience  data corruption.

## Expected Behavior
User should be able to read from cache the old data (data from before the corrupted IO was sent) or new data.

## Actual Behavior
User reads invalid data.

## Possible Fix
Set metadata status bits after data has been saved on disk.

mmichal10:
In WB mode metadata should be updated only if the actuall data had been saved
on disk. Otherwise metadata might be flushed to early which and consequently
data corruption might occur.

Fixes #519

### DIFF
--- a/src/engine/engine_wb.c
+++ b/src/engine/engine_wb.c
@@ -74,6 +74,8 @@ static int ocf_write_wb_do_flush_metadata(struct ocf_request *req)
 
 	env_atomic_set(&req->req_remaining, 1); /* One core IO */
 
+	_ocf_write_wb_update_bits(req);
+
 	if (req->info.flush_metadata) {
 		OCF_DEBUG_RQ(req, "Flush metadata");
 		ocf_metadata_flush_do_asynch(cache, req,
@@ -151,9 +153,6 @@ int ocf_write_wb_do(struct ocf_request *req)
 {
 	/* Get OCF request - increase reference counter */
 	ocf_req_get(req);
-
-	/* Update status bits */
-	_ocf_write_wb_update_bits(req);
 
 	/* Submit IO */
 	_ocf_write_wb_submit(req);
--- a/src/engine/engine_wt.c
+++ b/src/engine/engine_wt.c
@@ -110,8 +110,19 @@ static void _ocf_write_wt_req_complete(struct ocf_request *req)
 		req->complete(req, req->info.core_error ? req->error : 0);
 
 		ocf_engine_invalidate(req);
-	} else {
+
+		return;
+	}
+
+	if (req->info.dirty_any) {
+		/* Some of the request's cachelines changed its state to clean */
 		ocf_engine_push_req_front_if(req, &_io_if_wt_flush_metadata, true);
+	} else {
+		ocf_req_unlock_wr(ocf_cache_line_concurrency(req->cache), req);
+
+		req->complete(req, req->info.core_error ? req->error : 0);
+
+		ocf_req_put(req);
 	}
 }
 
@@ -163,6 +174,13 @@ static int _ocf_write_wt_do(struct ocf_request *req)
 {
 	/* Get OCF request - increase reference counter */
 	ocf_req_get(req);
+
+	if (!req->info.dirty_any) {
+		/* Set metadata bits before the request submission only if the dirty
+		   status for any of the request's cachelines won't change */
+		_ocf_write_wt_update_bits(req);
+		ENV_BUG_ON(req->info.flush_metadata);
+	}
 
 	/* Submit IO */
 	_ocf_write_wt_submit(req);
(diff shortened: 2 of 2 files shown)


src/engine/engine_wb.c
```diff
@@ -97,6 +97,9 @@ static void _ocf_write_wb_complete(struct ocf_request *req, int error)
 		req->error |= error;
 	}
 
+	if (req->error == 0)
```
arutk, line 100: this must be moved to after line 104 - otherwise you will flush status bits for all sectors before all data I/O is finished

src/engine/engine_wb.c
```diff
@@ -97,6 +97,9 @@ static void _ocf_write_wb_complete(struct ocf_request *req, int error)
 		req->error |= error;
 	}
```
arutk, line 98: Comments to commit message:
1. to -> too*
2. "which" seems to be out of place in this sentence: "Otherwise metadata might be flushed to early which and consequently data corruption might occur."

src/engine/engine_wb.c
mmichal10: done

src/engine/engine_wb.c
mmichal10: done

src/engine/engine_wb.c
```diff
@@ -109,6 +109,8 @@ static void _ocf_write_wb_complete(struct ocf_request *req, int error)
 
 		ocf_engine_invalidate(req);
 	} else {
+		_ocf_write_wb_update_bits(req);
```
arutk, line 112: sorry, you should actually move it to ocf_write_wb_do_flush_metadata - updating status bits acquires a mutex and here we are potentially in an atomic context.

robertbaldyga (dismissed):

arutk (approved):

michalwy (dismissed):
Similar change should be done in WT engine. There's a possibility that WT write is performed to dirty cache line (i.e. after switching WB->WT without flush) and status bits change from dirty to clean. If power failure occurs it might happen that recovery would ignore recent data from cache and would assume that data is clean while backend storage data is out of date.

src/engine/engine_wt.c
```diff
@@ -18,6 +18,84 @@
...
+	req->complete(req, req->info.core_error ? req->error : 0);
+
+	ocf_req_put(req);
+}
+
+static int ocf_write_wt_do_flush_metadata(struct ocf_request *req)
+{
+	struct ocf_cache *cache = req->cache;
+
+	env_atomic_set(&req->req_remaining, 1); /* One core IO */
```
robertbaldyga, line 78: This comment is misleading. This is actually refcount guard, it does not represent any IO.

robertbaldyga (approved):

src/engine/engine_wt.c
```diff
@@ -95,48 +159,11 @@ static inline void _ocf_write_wt_submit(struct ocf_request *req)
...
-		ocf_user_part_move(req);
-	}
-
-	ocf_hb_req_prot_unlock_wr(req);
-}
-
 static int _ocf_write_wt_do(struct ocf_request *req)
 {
 	/* Get OCF request - increase reference counter */
 	ocf_req_get(req);
```
arutk, line 166: We can detect the case where metadata flush is surely not necessary (!req->info.dirty_any) and set status bits here to avoid extra round trip via I/O queue

arutk: + add BUG_ON(!req->info.dirty_any && req->info.flush_metadata) after setting the bits here to make sure our assumption about metadata flush not being necessary holds.

arutk (changes requested):

src/engine/engine_wt.c
```diff
@@ -32,14 +110,18 @@ static void _ocf_write_wt_req_complete(struct ocf_request *req)
 		req->complete(req, req->info.core_error ? req->error : 0);
 
 		ocf_engine_invalidate(req);
+
+		return;
+	}
+
+	if (req->info.flush_metadata) {
```
arutk, line 117: I think this gets set in _ocf_write_wt_update_bit, so will always by false here. you should change this condition to req->info.dirty_any

src/engine/engine_wt.c
```diff
@@ -95,47 +170,17 @@ static inline void _ocf_write_wt_submit(struct ocf_request *req)
...
-
 static int _ocf_write_wt_do(struct ocf_request *req)
 {
 	/* Get OCF request - increase reference counter */
 	ocf_req_get(req);
 
-	/* Update status bits */
-	_ocf_write_wt_update_bits(req);
+	if (!req->info.dirty_any) {
+		/* Set metadata bits before the request submition only if the dirty
```
arutk, line 179: *submission

robertbaldyga (approved):

arutk (approved):

PR #525, opened 2021-06-25
Allow removing dirty core

mmichal10:


mmichal10:
Fixes https://github.com/Open-CAS/open-cas-linux/issues/844

PR #541, opened 2021-07-20
Sync cleaning deinit with running new iterations

mmichal10:


mmichal10:
Fixes https://github.com/Open-CAS/open-cas-linux/issues/886

src/cleaning/cleaning.c
```diff
@@ -49,6 +49,156 @@ struct cleaning_policy_ops cleaning_policy_ops[ocf_cleaning_max] = {
...
+}
+
+void ocf_cleaning_set_hot_cache_line(ocf_cache_t cache,
+		ocf_cache_line_t cache_line)
+{
+	ocf_cleaning_t policy = cache->conf_meta->cleaning_policy_type;
+
+	ENV_BUG_ON(policy >= ocf_cleaning_max);
+
+	if (!cleaning_policy_ops[policy].set_hot_cache_line)
```
jfckm, line 222: Since this is a hot IO path, and we're not using `nop` by default I'd consider marking this expression as unlikely

src/cleaning/cleaning.h
```diff
@@ -38,27 +38,6 @@ struct cleaning_policy_meta {
...
-	void (*set_hot_cache_line)(ocf_cache_t cache, uint32_t cache_line);
-	int (*set_cleaning_param)(ocf_cache_t cache, uint32_t param_id,
-			uint32_t param_value);
-	int (*get_cleaning_param)(ocf_cache_t cache, uint32_t param_id,
-			uint32_t *param_value);
-	void (*perform_cleaning)(ocf_cache_t cache, ocf_cleaner_end_t cmpl);
-	const char *name;
-};
-
-extern struct cleaning_policy_ops cleaning_policy_ops[ocf_cleaning_max];
```
jfckm, line 60: ❤️

robertbaldyga (changes requested):

src/cleaning/cleaning.c
```diff
@@ -49,6 +49,156 @@ struct cleaning_policy_ops cleaning_policy_ops[ocf_cleaning_max] = {
...
+
+	cleaning_policy_ops[policy].perform_cleaning(cache, cmpl);
+}
+
+const char *ocf_cleaning_get_name(ocf_cleaning_t policy)
+{
+	ENV_BUG_ON(!cleaning_policy_ops[policy].name);
+
+	return cleaning_policy_ops[policy].name;
+}
```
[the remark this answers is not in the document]
mmichal10: Making the functions static inline requires defining them in header file. If the definitions are in header file, `cleaning_policy_ops` array also has be visible in this header file. As a result pointers might be called directly, skipping the wrapper functions which is the opposite of what I want to achieve with this patch

src/cleaning/cleaning.c
```diff
@@ -49,6 +49,156 @@ struct cleaning_policy_ops cleaning_policy_ops[ocf_cleaning_max] = {
...
+
+void ocf_cleaning_deinitialize(ocf_cache_t cache, ocf_cleaning_t policy)
+{
+	ENV_BUG_ON(policy >= ocf_cleaning_max);
+
+	if (!cleaning_policy_ops[policy].deinitialize)
+		return;
+
+	cleaning_policy_ops[policy].deinitialize(cache);
+}
```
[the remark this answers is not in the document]
mmichal10: I've left `policy` param in init function because rollback procedure would become a bit tricky. In the current implementation we don't have to change `cache->conf_meta->cleaning_policy_typ` value until new cleaning policy is initialized.

src/mngt/ocf_mngt_flush.c
```diff
@@ -917,6 +917,18 @@ struct ocf_mngt_cache_set_cleaning_context
 	void *priv;
 };
 
+static void _ocf_mngt_cleaning_deinit_complete(void *priv)
+{
+	struct ocf_mngt_cache_set_cleaning_context *context = priv;
+	ocf_cache_t cache = context->cache;
+
+	ocf_refcnt_freeze(&cache->cleaner.cleaning_ctx);
```
robertbaldyga, line 925: At this point it's already frozen.

src/mngt/ocf_mngt_cache.c
```diff
@@ -215,7 +217,9 @@ static ocf_error_t __init_cleaning_policy(ocf_cache_t cache)
 
 static void __deinit_cleaning_policy(ocf_cache_t cache)
 {
+	ocf_refcnt_freeze(&cache->cleaner.cleaning_ctx);
 	ocf_cleaning_deinitialize(cache);
+	ocf_refcnt_unfreeze(&cache->cleaner.cleaning_ctx);
```
robertbaldyga: At this point there will be neither io in flight nor another mngt operation, so we can safely skip this.

src/cleaning/cleaning.h
```diff
@@ -40,6 +41,7 @@ struct cleaning_policy_meta {
 
 struct ocf_cleaner {
 	void *cleaning_policy_context;
+	struct ocf_refcnt cleaning_ctx __attribute__((aligned(64)));
```
robertbaldyga, line 44: This is wrong commit to introduce this field (it's used for the first time in the next commit), and I'd rather name it something like `refcnt` or `cleaning_refcnt`.

src/mngt/ocf_mngt_flush.c
```diff
@@ -917,6 +917,18 @@ struct ocf_mngt_cache_set_cleaning_context
...
+	ocf_cleaning_deinitialize(cache);
+	ocf_refcnt_unfreeze(&cache->cleaner.cleaning_ctx);
```
robertbaldyga, line 927: Don't we want to do unfreeze only after the new policy is initialized? Otherwise callbacks may be called before the new policy is initialized.

robertbaldyga (changes requested):

robertbaldyga (dismissed):

src/cleaning/cleaning.h
```diff
@@ -38,29 +39,9 @@ struct cleaning_policy_meta {
...
-			uint32_t *param_value);
-	void (*perform_cleaning)(ocf_cache_t cache, ocf_cleaner_end_t cmpl);
-	const char *name;
-};
-
-extern struct cleaning_policy_ops cleaning_policy_ops[ocf_cleaning_max];
-
 struct ocf_cleaner {
 	void *cleaning_policy_context;
+	struct ocf_refcnt refcnt __attribute__((aligned(64)));
```
mmkayPL, line 44: This isn't critical (since there aren't many instances of cleaner), but I suggest either sticking it at the beginning or find some space later in the structure - it's currently needlessly adding 2 cachelines worth of size, instead of one. Check
pahole -R modules/cas_cache/cas_cache.ko -I --hex

src/mngt/ocf_mngt_cache.c
```diff
@@ -294,6 +287,8 @@ static ocf_error_t init_attached_data_structures(ocf_cache_t cache)
 	__init_parts_attached(cache);
 	__populate_free(cache);
 
+	ocf_refcnt_init(&cache->cleaner.refcnt);
```
mmkayPL, line 290: Isn't this done in __init_cleaning_policy already?

jfckm (approved):

mmkayPL (approved):

robertbaldyga (approved):

PR #560, opened 2021-08-24
named anonymous struct - Golang compatibility issue with access to an…

KanagiAomori:
…onymous structs in open-cas-linux json API

Signed-off-by: Piotr Debski <piotr.debski@intel.com>


mmichal10:
Ok to test

mmichal10:
add to whitelist

inc/ocf_core.h
```diff
@@ -29,7 +29,7 @@ struct ocf_core_info {
 
 		/** Number of blocks left to flush in ongoing flush operation */
 		uint32_t dirty;
-	};
+	}flush_operation;
```
robertbaldyga, line 32: This unnecessarily breaks backwards compatibility. Why not just flatten this structure instead?

robertbaldyga (changes requested):
Please use shorter commit title (less than 80 chars is recommended). More detailed description can be included in the commit body.

robertbaldyga (approved):

PR #563, opened 2021-09-02
non-volatile metadata bug POC

beef9999:
API: `ocf_mngt_cache_load`

```
cache1: Loading Part config ERROR, invalid checksum
cache1: Metadata read FAILURE
cache1: Metadata Error
cache1: ERROR: Cannot load cache state
```

### DIFF
--- a/example/simple/src/volume.c
+++ b/example/simple/src/volume.c
@@ -7,8 +7,16 @@
 #include "volume.h"
 #include "data.h"
 #include "ctx.h"
+#include <unistd.h>
+#include <fcntl.h>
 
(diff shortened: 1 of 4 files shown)


example/simple/src/volume.c
```diff
@@ -7,8 +7,17 @@
 #include "volume.h"
 #include "data.h"
 #include "ctx.h"
+#include <unistd.h>
+#include <fcntl.h>
 
-#define VOL_SIZE 200*1024*1024
+const char* core_path = "/disk3/projects/cache-bench/example/core";
```
beef9999, line 13: Customize your path

PR #570, opened 2021-09-10
Fix attach with no cache_line_size specified

### LINKED ISSUE #564: [example bug] v21.3.3 example code is not working at all
1. The example code in v21.3.3 is not working at all. A floating point exception will occur, because the `ocf_mngt_cache_device_config_set_default` has set a zero cache-line size by default. 

2. Even after setting `device_cfg.cache_line_size = ocf_cache_line_size_default` correctly, the example code will still hang in the final stopping cache step.

3. v20.3.4 is OK

jfckm:
Fixes #564 

### DIFF
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -1036,7 +1036,8 @@ static void _ocf_mngt_init_properties(ocf_pipeline_t pipeline,
 
 	context->metadata.shutdown_status = ocf_metadata_clean_shutdown;
 	context->metadata.dirty_flushed = DIRTY_FLUSHED;
-	context->metadata.line_size = context->cfg.cache_line_size;
+	context->metadata.line_size = context->cfg.cache_line_size ?:
+			cache->metadata.line_size;
 
 	ocf_pipeline_next(pipeline);
 }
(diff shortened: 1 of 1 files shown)


rafalste (dismissed):

src/mngt/ocf_mngt_cache.c
```diff
@@ -2193,7 +2191,8 @@ static void _ocf_mngt_activate_init_properties(ocf_pipeline_t pipeline,
 
 	context->metadata.shutdown_status = ocf_metadata_dirty_shutdown;
 	context->metadata.dirty_flushed = DIRTY_NOT_FLUSHED;
-	context->metadata.line_size = context->cfg.cache_line_size;
+	context->metadata.line_size = context->cfg.cache_line_size ?:
```
robertbaldyga, line 2194: In activate path it actually should not be taken from cfg in any case.

src/mngt/ocf_mngt_cache.c
```diff
@@ -2734,20 +2732,28 @@ void ocf_mngt_cache_activate(ocf_cache_t cache,
...
 		OCF_CMPL_RET(cache, priv, -OCF_ERR_INVAL);
+	}
 
-	/* Activate is not allowed with 'force' flag on */
-	if (cfg->force) {
-		ocf_cache_log(cache, log_err, "Using 'force' flag is forbidden "
-				"for activate operation.");
+	if (cache->metadata.is_volatile) {
+		ocf_cache_log(cache, log_err, "Activating volatile cache is"
+				"forbidden.");
```
robertbaldyga, line 2750: This check is already done bind, so no need for doing it there.

robertbaldyga (changes requested):

robertbaldyga (approved):

arutk (approved):

PR #572, opened 2021-09-15
Non volatile metadata bug POC

beef9999:
#### API: `ocf_mngt_cache_load`

```
cache1: Loading Part config ERROR, invalid checksum
cache1: Metadata read FAILURE
cache1: Metadata Error
cache1: ERROR: Cannot load cache state
```

#### How to reproduce?

This POC is based on the example, it uses two files as cache and core device. `metadata_volatile` is set to false.

Then run the example twice, you will see the first time succeeded, while the second failed when trying to reload cache. 

Error message is as above.

### DIFF
--- a/example/simple/src/volume.c
+++ b/example/simple/src/volume.c
@@ -7,8 +7,16 @@
 #include "volume.h"
 #include "data.h"
 #include "ctx.h"
+#include <unistd.h>
+#include <fcntl.h>
 
 #define VOL_SIZE 200*1024*1024
+
+const char* core_path = "core";
+const char* cache_path = "cache";
+
+int core_fd = 0;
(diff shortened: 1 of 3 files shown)


example/simple/src/volume.c
```diff
@@ -7,9 +7,17 @@
 #include "volume.h"
 #include "data.h"
 #include "ctx.h"
+#include <unistd.h>
+#include <fcntl.h>
 
 #define VOL_SIZE 200*1024*1024
 
+const char* core_path = "/disk3/projects/cache-bench/example/core";
+const char* cache_path = "/disk3/projects/cache-bench/example/cache";
```
beef9999, line 16: customize your file path

PR #573, opened 2021-09-15
Fix io-deadlock when stopping cache with synchronous queue

jfckm:


arutk:
My opinion is that solving his kind of problems once and for all requires OCF adapters to always implement asynchronous kick and enforce async kick for pipeline steps.

beef9999:
> My opinion is that solving his kind of problems once and for all requires OCF adapters to always implement asynchronous kick and enforcing async kick for pipeline steps.
> ... and adding the sample compilation and execution to CI :)

Agreed. Then sample code will need a new async engine.
How about transplant the sample code based on SPDK? And leave it to users to setup SPDK environment?

arutk:
issue will be fixed by updating the example code

PR #575, opened 2021-09-17
simple example: implement queues based on threads

### LINKED ISSUE #564: [example bug] v21.3.3 example code is not working at all

arutk:
Fixes #564 

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/575)

### DIFF
--- a/example/simple/src/main.c
+++ b/example/simple/src/main.c
@@ -7,9 +7,11 @@
 #include <stdlib.h>
 #include <errno.h>
 #include <string.h>
+#include <semaphore.h>
 #include <ocf/ocf.h>
 #include "data.h"
 #include "ctx.h"
+#include "queue_thread.h"
 
 /*
  * Cache private data. Used to share information between async contexts.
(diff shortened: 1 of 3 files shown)


beef9999:
Which C++ version does it require, 20 ?

beef9999:
I was thinking to suggest pure async C libraries like State Threads, but indeed C++11 threads is a much better solution as a sample code.

Splendid.

arutk:
Hi @beef9999. Thanks for your feedback. Unfortunately after some more consideration I have realized, that mixing pthread usage with c++ threads is potentially unsafe. And our OCF enviromnent for the sample is based on the pthread. So I should probably change this PR to use pthreads instead of c++ standard library :(.

arutk:
@beef9999 sure, I was about to debug your POC. will let you know

beef9999 (approved):

example/simple/src/queue_thread.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	int ret = 0;
+
+	try {
+		struct queue_thread* mngt_queue_thread = new queue_thread(mngt_queue);
+		struct queue_thread* io_queue_thread = new queue_thread(io_queue);
+
+		ocf_queue_set_priv(mngt_queue, mngt_queue_thread);
+		ocf_queue_set_priv(io_queue, io_queue_thread);
+	} catch(...) {
```
kmajzero, line 70: Pokemon-style exception handling does not look good. How about "catch(std::bad_alloc)"  ;) ?

arutk: I have changed implementation to C

example/simple/src/queue_thread.c
```diff
@@ -0,0 +1,152 @@
...
+			qt->signalled = false;
+			pthread_mutex_unlock(&qt->mutex);
+
+			/* execute items on the queue */
+			ocf_queue_run(q);
+
+			pthread_mutex_lock(&qt->mutex);
+		}
+
+		pthread_cond_wait(&qt->cv, &qt->mutex);
```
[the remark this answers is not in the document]
arutk: looks like wait is unnecessary in this case

example/simple/src/main.c
```diff
@@ -157,8 +141,13 @@ int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
...
 
+	ret = initialize_threads(cache_priv->mngt_queue, cache_priv->io_queue);
+	if (ret)
+		goto err_cache;
+
 	/* Attach volume to cache */
 	ocf_mngt_cache_attach(*cache, &device_cfg, simple_complete, &context);
+	sem_wait(&context.sem);
 	if (ret)
 		goto err_cache;
```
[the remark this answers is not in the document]
arutk: ret is referenced via pointer in simple_complete :O

example/simple/src/queue_thread.c
```diff
@@ -0,0 +1,152 @@
...
+
+			/* execute items on the queue */
+			ocf_queue_run(q);
+
+			pthread_mutex_lock(&qt->mutex);
+		}
+
+		if (!qt->stop)
+			pthread_cond_wait(&qt->cv, &qt->mutex);
+	} while (!qt->stop);
```
rafalste, line 111: Maybe this will do the job:
```
while (!qt->stop && qt->signalled) {
	qt->signalled = false;
	pthread_mutex_unlock(&qt->mutex);

	/* execute items on the queue */
	ocf_queue_run(q);

	pthread_mutex_lock(&qt->mutex);

	pthread_cond_wait(&qt->cv, &qt->mutex);
}
```

arutk: Looks good. I just needed to add an extra condition before pthread_cond_wait to make sure we don't wait with signalled or stop set.

example/simple/src/queue_thread.c
```diff
@@ -0,0 +1,152 @@
...
+
+			/* execute items on the queue */
+			ocf_queue_run(q);
+
+			pthread_mutex_lock(&qt->mutex);
+		}
+
+		if (!qt->stop && !qt->signalled) 
+			pthread_cond_wait(&qt->cv, &qt->mutex);
+	} while (!qt->stop);
```
rafalste, line 111: Leftover do-while loop condition.

rafalste (approved):

beef9999:
Can this patch be merged now?

beef9999:
@arutk After some careful thought, I now suspect this queue implementation was done in with a wrong way that has reduced the I/O currency to 1. User may submit I/O concurrently, but this while{} loop with signal notification is not parallel.

I think start a temporary new thread and run ocf_queue_run is just enough.

beef9999:
In my own implementation, I put the condition variable into the `volume_data` and trigger it in `read_complete` or `write_complete`. Looks like it's working and achieved good performance. Don't know if it is flawed.

rafalste (approved):

arutk:
Hi @beef9999. It is true that the thread loop is not parallelized itself. But this is our intention. Queue is an abstraction for a unit of concurrency. In actual product you would have multiple queues and parallelization would be achieved by running them in parallel.

kmajzero (approved):

PR #576, opened 2021-09-17
Always use async queue kick in management pipeline

arutk:
Management pipelines tend to consist of multiple asynchronous steps.
Allowing synchronous queue kick results in massive call stacks (e.g.
almost 500 functions deep in case of cache stop). Since async kick
is required anyway, it seems reasonable to switch to async kick
in pipeline implementation.


beef9999:
I was thinking to suggest pure async C libraries like State Threads, but indeed C++11 threads is a much better solution as a sample code.

Splendid.

robertbaldyga (approved):

rafalste (approved):

PR #579, opened 2021-10-05
Maintaining runtime metadata structures in failover standby

mmichal10:
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/579)


src/ocf_cache_priv.h
```diff
@@ -164,7 +164,7 @@ static inline uint64_t ocf_get_cache_occupancy(ocf_cache_t cache)
 	ocf_core_t core;
 	ocf_core_id_t core_id;
 
-	for_each_core(cache, core, core_id)
+	for_each_core_all(cache, core, core_id)
```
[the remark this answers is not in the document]
mmichal10: When a core is being removed all the associated cachelines are invalidated so occupancy stats has to match. Detach doesn't change value of `core->added` so both `for_each_core()` and `for_each_core_all()` macros won't skip detached core. 

`Imo it would be safer to not have any assumptions about data structures of invalid cores.`
I agree it would be safer but without these assumptions we won't be able to perform recovery during passive mode. Each recovered cache lines has to be assigned to a core. And since we don't have access to valid superblock before activate we have to assume that any core my be valid. Otherwise assigning cache lines to cores has to be done during  the activate which is time consuming.

src/mngt/ocf_mngt_cache.c
```diff
@@ -1991,6 +2006,18 @@ static void _ocf_mngt_bind_init_attached_structures(ocf_pipeline_t pipeline,
...
 }
 
+static void _ocf_mngt_bind_recovery_unsafe(ocf_pipeline_t pipeline,
+		void *priv, ocf_pipeline_arg_t arg)
+{
+	struct ocf_cache_attach_context *context = priv;
+	ocf_cache_t cache = context->cache;
+
+	_ocf_mngt_bind_rebuild_metadata(cache);
+	__populate_free(cache);
```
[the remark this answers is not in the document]
mmichal10: `_ocf_mngt_load_post_metadata_load()` isn't a step of the activate anymore

src/cleaning/acp.c
```diff
@@ -614,6 +614,9 @@ void cleaning_policy_acp_set_hot_cache_line(struct ocf_cache *cache,
 	struct acp_cleaning_policy_meta *acp_meta;
 	struct acp_chunk_info *chunk;
 
+	if (unlikely(!ocf_cache_is_running(cache)))
```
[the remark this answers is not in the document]
mmichal10: `set_hot` is NOP only for acp, alru needs to dirty cachelines all the time. If we're fine with the additional condition `if (unlikely(!ocf_cache_is_running(cache)) && cache->cleaning_policy == acp)` as a step of `set_hot()` wrapper I can change it, but imo both ways seems similar.

src/mngt/ocf_mngt_cache.c
```diff
@@ -2008,6 +2008,8 @@ static void _ocf_mngt_load_unsafe_complete(void *priv, int error)
 				-OCF_ERR_START_CACHE_FAIL);
 	}
 
+	context->metadata.shutdown_status = ocf_metadata_dirty_shutdown;
```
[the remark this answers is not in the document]
mmichal10: That's true. Do you think this value should be set is some other place in code?

src/mngt/ocf_mngt_cache.c
```diff
@@ -2335,6 +2365,7 @@ struct ocf_pipeline_properties _ocf_mngt_cache_activate_pipeline_properties = {
 		OCF_PL_STEP(_ocf_mngt_test_volume),
 		OCF_PL_STEP(_ocf_mngt_init_promotion),
 		OCF_PL_STEP(_ocf_mngt_load_add_cores),
+		OCF_PL_STEP(_ocf_mngt_activate_cleaning_add_cores),
```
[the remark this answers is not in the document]
mmichal10: During the regular recovery adding core to the cleaning policy is a step of reinitializing cleaning policy. It can be done because we have access to the valid superblock and info about added cores.
Missing info about valid cores is the reason why cores can't be added to the cleaning policy during the passive recovery and needs to be postponed to the activate.

src/metadata/metadata.c
```diff
@@ -2045,6 +2045,162 @@ void ocf_metadata_probe_cores(ocf_ctx_t ctx, ocf_volume_t volume,
...
+		ocf_cache_line_t cline)
+{
+	ocf_core_id_t core_id;
+	uint64_t core_line;
+
+	ocf_metadata_get_core_info(cache, cline, &core_id, &core_line);
+
+	if (metadata_test_dirty(cache, cline) && core_id < OCF_CORE_MAX) {
+		/* Cline stays on the default partition*/
+		ENV_BUG_ON(!metadata_test_valid_any(cache, cline));
```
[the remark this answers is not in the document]
mmichal10: It's not the first bug_on in IO path. At this point metadata which was both dirty and invalid should be already handled and if it's not, CAS shouldn't operate anymore.

src/metadata/metadata.c
```diff
@@ -2045,6 +2045,162 @@ void ocf_metadata_probe_cores(ocf_ctx_t ctx, ocf_volume_t volume,
...
 
+static inline void _reset_cline(ocf_cache_t cache, ocf_cache_line_t cline)
+{
+	/* The cacheline used to be dirty but it is not anymore so it needs to be
+	   moved to a clean lru list */
+	ocf_lru_clean_cline(cache, &cache->user_parts[PARTITION_DEFAULT].part,
+			cline);
+
+	ocf_lru_rm_cline(cache, cline);
+	ocf_metadata_set_partition_id(cache, cline, PARTITION_FREELIST);
```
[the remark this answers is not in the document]
mmichal10: `ocf_lru_rm_cline()` moves cacheline to `cache->free`

src/metadata/metadata.c
```diff
@@ -1566,6 +1566,31 @@ bool ocf_metadata_clear_valid_if_clean(struct ocf_cache *cache,
 	}
 }
 
+void ocf_metadata_amend(struct ocf_cache *cache,
```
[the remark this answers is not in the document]
mmichal10: done

src/mngt/ocf_mngt_cache.c
```diff
@@ -541,31 +541,58 @@ static void _ocf_mngt_bind_rebuild_metadata(ocf_cache_t cache)
...
+
+	result = _ocf_init_cleaning_policy(cache,
+			cache->conf_meta->cleaning_policy_type,
+			context->metadata.shutdown_status);
+
+	if (result)
+		OCF_PL_FINISH_RET(pipeline, result);
+
+	ocf_pipeline_next(pipeline);
+}
```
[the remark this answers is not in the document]
mmichal10: done

src/metadata/metadata.c
```diff
@@ -2045,6 +2045,162 @@ void ocf_metadata_probe_cores(ocf_ctx_t ctx, ocf_volume_t volume,
...
+
+static void cleanup_old_mapping(ocf_cache_t cache, ocf_cache_line_t begin,
+		ocf_cache_line_t end)
+{
+	ocf_cache_line_t cline;
+
+	/* The last page of collision section may contain fewer entries than
+	   entries_in_page */
+	if (end > ocf_metadata_collision_table_entries(cache))
+		end = ocf_metadata_collision_table_entries(cache);
```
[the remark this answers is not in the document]
mmichal10: done

src/ocf_cache_priv.h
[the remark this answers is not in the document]
robertbaldyga: In standby mode the cumulative occupancy is required only for `ocf_lru_populate()` which uses this value only for sanity check, so this is not a strong dependency. How about disabling this sanity check in standby mode?

src/mngt/ocf_mngt_cache.c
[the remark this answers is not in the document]
robertbaldyga: In this particular commit it is.

src/mngt/ocf_mngt_cache.c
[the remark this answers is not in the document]
robertbaldyga: In standby pipeline we call `_ocf_mngt_init_properties()` which sets shutdown status to clean. Maybe we should replace this step with standby-specific one, that sets it already to dirty instead of toggling this status every second step ;)

arutk:
I suggest adding some description to the first ("Improve metadata passive update") commit

arutk:
Also please make the title of this PR more descriptive. Maybe something along the lines of "maintaining runtime metadata structures in failover standby"

src/mngt/ocf_mngt_cache.c
```diff
@@ -2183,11 +2183,15 @@ static void _ocf_mngt_activate_check_superblock_complete(void *priv, int error)
 		OCF_PL_FINISH_RET(context->pipeline, result);
 
 	if (cache->conf_meta->metadata_layout != cache->metadata.layout) {
+		ocf_cache_log(cache, log_err, "Failed to activate failover instance: "
```
arutk, line 2261: I think "standby instance" or "failover standby instance" is more precise than "failover instance" in this case

src/utils/utils_cache_line.c
```diff
@@ -167,3 +167,40 @@ void set_cache_line_dirty(struct ocf_cache *cache, uint8_t start_bit,
 
 	ocf_cleaning_set_hot_cache_line(cache, line);
 }
+
+void cline_rebuild_metadata(ocf_cache_t cache,
```
arutk, line 171: I think our convention is to always have externally linked function names prefixed with "ocf_"

src/cleaning/cleaning.h
```diff
@@ -41,6 +41,7 @@ struct cleaning_policy_meta {
 
 struct ocf_cleaner {
 	struct ocf_refcnt refcnt __attribute__((aligned(64)));
+	ocf_cleaning_t policy;
```
arutk, line 44: I suggest rephrasing commit title: "store cleaning policy info in additional location outside superblock"

src/ocf_lru.c
```diff
@@ -194,7 +194,7 @@ static void remove_lru_list_nobalance(ocf_cache_t cache,
 	node->hot = false;
 }
 
-static void remove_lru_list(ocf_cache_t cache, struct ocf_lru_list *list,
+void remove_lru_list(ocf_cache_t cache, struct ocf_lru_list *list,
```
arutk, line 197: I would rename this to "ocf_lru_remove_locked" since this is now part of LRU list api.

src/concurrency/ocf_pio_concurrency.c
```diff
@@ -0,0 +1,220 @@
+/*
+ * Copyright(c) 2021 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause-Clear
```
arutk, line 3: remove "clear"

src/concurrency/ocf_pio_concurrency.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright(c) 2021-2021 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause-Clear
```
arutk, line 3: remove "clear"

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,369 @@
+/*
+ * Copyright(c) 2012-2021 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause-Clear
```
arutk, line 3: "remove clear"

src/metadata/metadata_passive_io.h
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright(c) 2012-2021 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause-Clear
```
arutk, line 3: remove "clear"

arutk (changes requested):

src/cleaning/cleaning.h
mmichal10: How about "Additional cleaning policy info outside of the SB"? It fits the 50 characters limit.

src/mngt/ocf_mngt_cache.c
```diff
@@ -493,39 +493,54 @@ static void _recovery_reset_cline_metadata(struct ocf_cache *cache,
...
+}
+
+static void _ocf_mngt_bind_rebuild_metadata(ocf_cache_t cache)
+{
+	ocf_cleaning_t clean_policy = cache->conf_meta->cleaning_policy_type;
+	cache->conf_meta->cleaning_policy_type = ocf_cleaning_nop;
+
+	_ocf_mngt_rebuild_metadata(cache, true);
+
+	cache->conf_meta->cleaning_policy_type = clean_policy;
```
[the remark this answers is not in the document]
mmichal10: Keeping additional copy of cleaning policy outside of the superblock allowed to remove setting `nop` here

src/ocf_cache_priv.h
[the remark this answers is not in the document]
mmichal10: I've added `safe` param to the `ocf_lru_populate()`. Let me know what do you think about the new way of populating the freelist during the standby-init-recovery

robertbaldyga (changes requested):

robertbaldyga (changes requested):

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,369 @@
...
+	int result;
+	char name[MPOOL_ALLOCATOR_NAME_MAX];
+	uint32_t max_io_size;
+
+	result = snprintf(name, sizeof(name), "ocf_cache_pio");
+	if (result < 0 || result >= sizeof(name))
+		return result;
+
+	max_io_size = ocf_volume_get_max_io_size(ocf_cache_get_volume(cache));
+	cache->passive.allocator = env_allocator_create(max_io_size, name, true);
```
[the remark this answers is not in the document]
robertbaldyga: Actually it should be `sizeof(struct ocf_request) + OCF_DIV_ROUND_UP(BYTES_TO_PAGES(max_io_size), 8)`

src/mngt/ocf_mngt_cache.c
```diff
@@ -533,31 +533,73 @@ static void _ocf_mngt_rebuild_metadata(ocf_cache_t cache, bool fix_metadata)
...
 
 	if (context->metadata.shutdown_status != ocf_metadata_clean_shutdown) {
 		_ocf_mngt_recovery_rebuild_metadata(cache);
 		__populate_free_safe(cache);
 	}
 
-	cleaning_policy = cache->conf_meta->cleaning_policy_type;
+	result = _ocf_init_cleaning_policy(cache,
+			cache->conf_meta->cleaning_policy_type,
+			context->metadata.shutdown_status);
```
arutk: Since you have added a separate pipeline step "_ocf_mngt_activate_init_cleaning", maybe you can use it in load pipeline as a step after _ocf_mngt_load_post_metadata_load

src/mngt/ocf_mngt_cache.c
```diff
@@ -565,23 +565,6 @@ static inline ocf_error_t _ocf_init_cleaning_policy(ocf_cache_t cache,
 	return result;
 }
 
-static void _ocf_mngt_activate_init_cleaning(ocf_pipeline_t pipeline,
```
arutk: you have a typo in commit message "cachace insance"

src/mngt/ocf_mngt_common.c
```diff
@@ -151,6 +151,13 @@ void ocf_mngt_cache_put(ocf_cache_t cache)
 	}
 }
 
+void __set_cleaning_policy(ocf_cache_t cache,
```
arutk, line 154: suggest renaming to ocf_set_cleaning_policy

src/metadata/metadata.c
```diff
@@ -18,6 +18,8 @@
 #include "../utils/utils_cache_line.h"
 #include "../utils/utils_io.h"
 #include "../utils/utils_pipeline.h"
+#include "../concurrency/ocf_pio_concurrency.h"
+#include "../engine/engine_common.h"
```
arutk, line 22: is this one needed?

src/concurrency/ocf_pio_concurrency.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright(c) 2021-2021 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef OCF_PIO_CONCURRENCY_H_
```
arutk, line 6: I think we typically prefix these preprocesor definitions with one or two "_"

arutk (changes requested):

src/mngt/ocf_mngt_common.c
mmichal10: `ocf_set_cleaning_policy` may suggest the function responsible for switching cleaning policy whereas it's only a helper function. Double underscore and the lack of `ocf` prefix are supposed to highlight it.

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,377 @@
...
+	.read = passive_io_resume,
+	.write = passive_io_resume,
+};
+
+static void passive_io_page_lock_acquired(struct ocf_request *req)
+{
+	ocf_engine_push_req_front(req, true);
+}
+
+int ocf_metadata_passive_update(ocf_cache_t cache, struct ocf_io *io,
```
arutk, line 289: I think the file name is misleading, there is no I/O in this file (metadata_passive_io.c)

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,377 @@
...
+		return 0;
+	}
+
+	if (io_start_page >= ctrl->count_pages) {
+		io_cmpl(io, 0);
+		return 0;
+	}
+
+	if (io->addr % PAGE_SIZE || io->bytes % PAGE_SIZE) {
+		ocf_cache_log(cache, log_crit,
```
arutk, line 309: imo it is a warning level severity, not critical

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	req->io_queue = io->io_queue;;
+	req->info.internal = true;
+	req->io_if = &passive_io_restart_if;
+	req->rw = OCF_WRITE;
+	req->data = io;
+	req->master_io_req = io_cmpl;
+	req->cache = cache;
+	env_atomic_set(&req->lock_remaining, 0);
```
arutk, line 336: extracting generic part (required for alock) from ocf_request structure and creating dedicated passive request struct would make this look way better

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,377 @@
...
+
+	req->core_line_first = io_start_page;
+	req->core_line_count = io_end_page - io_start_page;
```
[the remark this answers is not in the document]
arutk: my comment for passive_io_resume applies here as well: I would make the "_end" variable actually store the index of end page and then 
req->core_line_count = io_end_page - io_start_page + 1;

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,377 @@
...
+}
+
+static void handle_previously_invalid(ocf_cache_t cache,
+		ocf_cache_line_t cline)
+{
+	ocf_core_id_t core_id;
+	uint64_t core_line;
+	uint32_t lock_idx = ocf_metadata_concurrency_next_idx(cache->mngt_queue);
+
+	ocf_metadata_get_core_info(cache, cline, &core_id, &core_line);
```
arutk, line 78: Can you please add a comment here explaining, that cacheline mapping is fixed at the moment thanks to pio lock being held? This is another reason to have pio locks, let's try to document this somehow.

arutk: ... also status bits are fixed for the same reason (next line - metadata_test_dirty())

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,377 @@
...
+		OCF_METADATA_LRU_WR_LOCK(cline);
+		ocf_cline_rebuild_metadata(cache, core_id, core_line, cline);
+		OCF_METADATA_LRU_WR_UNLOCK(cline);
+		ocf_hb_cline_prot_unlock_wr(&cache->metadata.lock, lock_idx, core_id,
+				core_line);
+		ocf_cleaning_set_hot_cache_line(cache, cline);
+
+	} else {
+		/* Moving cline from the default partition to the freelist */
+		ocf_cleaning_purge_cache_block(cache, cline);
```
arutk, line 130: ocf_cleaning_purge_cache_block attempts to increment cache->cleaner.refcnt - just double checking to make sure this isn't frozen in standby

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,377 @@
...
+	}
+}
+
+static void _dec_core_stats(ocf_core_t core)
+{
+	ocf_part_id_t part = PARTITION_DEFAULT;
+
+	ENV_BUG_ON(env_atomic_dec_return(&core->runtime_meta->cached_clines) < 0);
+	ENV_BUG_ON(env_atomic_dec_return(
+				&core->runtime_meta->part_counters[part].cached_clines) < 0);
```
arutk, line 173: lines too long

arutk (changes requested):

src/metadata/metadata_passive_io.c
mmichal10: It's probably caused by tab width equal to 8 :) with vim configured to `tabstop=4` it fits 80 characters limit

src/metadata/metadata_passive_io.c
mmichal10: The counter is frozen only to switch active cleaning policy which is a disabled operation in standby mode

src/metadata/metadata_passive_io.c
```diff
@@ -0,0 +1,377 @@
...
+		metadata_segment_part_config,
+		metadata_segment_core_config,
+		metadata_segment_core_uuid,
+		metadata_segment_collision,
+	};
+	int i;
+
+	for (i = 0; i < ARRAY_SIZE(update_segments); i++) {
+		ocf_cache_line_t cache_line_range_end, cache_line_range_start;
+		enum ocf_metadata_segment_id seg = update_segments[i];
```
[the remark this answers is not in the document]
mmichal10: The reason why cache lines range is calculated here, not in `cleanup_old_mapping()` and `update_list_segment()` is that I wanted to avoid passing `struct ocf_metadata_raw` to them. With the current implementation both functions operate only in a domain of cache lines and I prefer not to change it. I will move calculating the proper size of the range to the main function and pass it instead of end of range

src/concurrency/ocf_pio_concurrency.c
```diff
@@ -0,0 +1,228 @@
...
+		pio_ctx->segments[i].id = update_segments[i];
+		pio_ctx->segments[i].begin = raw->ssd_pages_offset;
+		pio_ctx->segments[i].end = raw->ssd_pages_offset + raw->ssd_pages;
+
+		pages_to_alloc += raw->ssd_pages;
+	}
+
+	ret = ocf_alock_init_inplace(alock, pages_to_alloc, name, &ocf_pio_conc_cbs, cache);
+	if (ret) {
+		env_free(alock);
```
arutk, line 217: env_vfree

src/concurrency/ocf_pio_concurrency.c
```diff
@@ -0,0 +1,228 @@
...
+	uint32_t pages_to_alloc = 0;
+	enum ocf_metadata_segment_id update_segments[] = {
+		metadata_segment_sb_config,
+		metadata_segment_part_config,
+		metadata_segment_core_config,
+		metadata_segment_core_uuid,
+		metadata_segment_collision,
+	};
+	int i;
+
```
arutk, line 190: maybe add BUG_ON(ARRAY_SIZE(update_segments) > ARRAY_SIZE(pio_ctx->segments))

arutk (approved):

src/metadata/metadata_passive_io.c
mmichal10: I renamed it to metadata_passive_update.c

src/concurrency/ocf_pio_concurrency.c
mmichal10: I added build bugon

robertbaldyga: I'd also add "FIXME: pio_concurrency should be parametrizable with segments list", to not leave this BUG_ON() as a lone monument of poor design.

robertbaldyga (approved):

arutk (approved):

PR #606, opened 2021-12-10
[21.3.4]Backport recent fixes

mmichal10:


rafalste (dismissed):

arutk (approved):

src/ocf_core.c
```diff
@@ -341,14 +341,17 @@ int ocf_core_submit_io_fast(struct ocf_io *io)
...
 	fast = ocf_engine_hndl_fast_req(req);
 	if (fast != OCF_FAST_PATH_NO) {
 		ocf_trace_push(io->io_queue, &trace_event, sizeof(trace_event));
 		ocf_core_seq_cutoff_update(core, req);
+		ocf_req_put(req);
 		return 0;
 	}
 
+	ocf_req_put(req);
 	dec_counter_if_req_was_dirty(req);
```
arutk, line 355: shouldn't this put go after dec_counter_if_req_was_dirty?

src/ocf_core.c
```diff
@@ -341,14 +341,17 @@ int ocf_core_submit_io_fast(struct ocf_io *io)
...
 
 	ocf_io_put(io);
```
arutk, line 357: if I recall correctly io allocation actualy a part of req allocation. So using io_get/io_put here is misleading. I would remove them entirely, or at least move io_put() before "req_put()" with an appropriate comment

arutk (approved):

src/ocf_core.c
robertbaldyga: Because of this `ocf_get_io()/ocf_put_io()` surrounding call to engine, there is guarantee that request will not be freed, so this fix is not needed on 21.3 branch at all.

arutk (approved):

robertbaldyga (approved):

PR #611, opened 2021-12-15
pyocf: core management operations power failure recovery test

arutk:


tests/functional/tests/surprise_shutdown/test_management_suprise_shutdown.py
```diff
@@ -0,0 +1,147 @@
...
+    def tested_func(cache):
+        cache.remove_core(core)
+
+    def check_func(cache, error_triggered):
+        stats = cache.get_stats()
+        assert stats["conf"]["core_count"] == (1 if error_triggered else 0)
+
+    mngmt_op_surprise_shutdown_test(pyocf_ctx, tested_func, prepare_func, check_func)
+
+
```
arutk, line 108: missing test for double I/O error in management operation - to detect whether flapping idx is corretly reverted in case of flush pipeline error

PR #616, opened 2021-12-19
Extend cache recovery collision metadata validation 

arutk:
In current design there might be residual mappings in collision segment . Need to invalidate all mappings associated with cores not currently added to the cache instance. Additionally fail recovery if status bits are inconsistent (dirty & !valid).


src/metadata/metadata.c
```diff
@@ -1204,10 +1204,13 @@ static void _recovery_reset_cline_metadata(struct ocf_cache *cache,
 	ocf_metadata_set_core_info(cache, cline, OCF_CORE_MAX, ULLONG_MAX);
 
 	metadata_clear_valid(cache, cline);
+	metadata_clear_dirty(cache, cline);
```
mmichal10, line 1207: You can use `metadata_init_status_bits()` instead

mmichal10 (approved):

src/metadata/metadata.c
arutk: it's done, thanks

src/metadata/metadata.c
```diff
@@ -1225,7 +1225,13 @@ static void _recovery_rebuild_metadata(ocf_pipeline_t pipeline,
 
 	for (cline = 0; cline < collision_table_entries; cline++) {
 		ocf_metadata_get_core_info(cache, cline, &core_id, &core_line);
+		if (!ocf_metadata_check(cache, cline) ||
+				core_id > OCF_CORE_MAX) {
```
jfckm, line 1229: should `OCF_CORE_MAX` be included as a valid value? This value is also `OCF_CORE_ID_INVALID`. Maybe you meant `OCF_CORE_ID_MAX`?

arutk: OCF_CORE_MAX is a special value indicating unused mapping and should not result in cache recovery error

jfckm (approved):

mmichal10 (approved):

PR #630, opened 2021-12-23
atomic initialization of on-disk metadata during cache attach

arutk:


mmichal10 (dismissed):

jfckm (approved):

src/metadata/metadata_raw.c
```diff
@@ -645,6 +697,7 @@ static const struct raw_iface IRAW[metadata_raw_type_max] = {
 		.page			= _raw_ram_page,
 		.access			= _raw_ram_access,
 		.update			= _raw_ram_update,
+		.zero			= NULL,
```
robertbaldyga, line 700: I believe it should be `raw_ram_zero()` instead of `NULL`. Otherwise we have nullptr dereference in attach for attomic mode. On top of that the configuration metadata in atomic mode is handled the same way as in normal, so without zeroing the superblock we'd have the same problem with potential metadata inconsistency.

robertbaldyga (changes requested):

src/metadata/metadata_superblock.c
```diff
@@ -438,6 +438,8 @@ struct ocf_metadata_superblock
 {
 	struct ocf_metadata_segment segment;
 	struct ocf_superblock_config *config;
+	struct ocf_metadata_raw *raw;
+	ocf_cache_t cache;
```
robertbaldyga, line 442: Why do we need those? The `ocf_metadata_sb_zero()` could just take those as args, as those are available in the call context anyway.

robertbaldyga (changes requested):

src/metadata/metadata_superblock.c
arutk: to make it a self contained object

src/metadata/metadata_raw.c
arutk: ok

src/mngt/ocf_mngt_cache.c
```diff
@@ -2139,6 +2164,7 @@ struct ocf_pipeline_properties _ocf_mngt_cache_standby_attach_pipeline_propertie
 		OCF_PL_STEP(_ocf_mngt_standby_init_cleaning),
 		OCF_PL_STEP(_ocf_mngt_standby_preapre_mempool),
 		OCF_PL_STEP(_ocf_mngt_standby_init_pio_concurrency),
+		OCF_PL_STEP(_ocf_mngt_zero_superblock),
```
arutk, line 2168: added after rebase

robertbaldyga (approved):

mmichal10 (approved):

PR #639, opened 2022-01-13
Fix error handling in cache attach

arutk:
Only close cores in error handling if attach parameter "open_cores" is
set to true.

### DIFF
--- a/src/mngt/ocf_mngt_cache.c
+++ b/src/mngt/ocf_mngt_cache.c
@@ -340,8 +340,9 @@ static void init_attached_data_structures_recovery(ocf_cache_t cache,
  * Used in case of cache initialization errors.			*
  ****************************************************************/
 static void _ocf_mngt_close_all_uninitialized_cores(
-		ocf_cache_t cache)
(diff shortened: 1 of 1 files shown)


robertbaldyga (approved):

src/mngt/ocf_mngt_cache.c
```diff
@@ -350,7 +351,8 @@ static void _ocf_mngt_close_all_uninitialized_cores(
 			continue;
 
 		volume = &(cache->core[i].volume);
-		ocf_volume_close(volume);
+		if (context->cfg.open_cores)
+			ocf_volume_close(volume);
```
rafalste, line 355: Since *volume* is used only for closing, maybe you can also initialize it inside this *if* condition?

arutk: I like it more this way, less lines of code and it's easier to read for me. Compiler will most likely optimize the initialization for us.

rafalste (approved):

pdebski21 (approved):

PR #641, opened 2022-01-16
utils: call io_put before completion function

hammerg:
In the simple example, after calling ocf_mngt_cache_stop, the program
never exit while it waits endlessly for the volume's reference count
to reach zero. Changing the functions call order reduce the reference
count before the flush continue its work.

### DIFF
--- a/src/utils/utils_io.c
+++ b/src/utils/utils_io.c
@@ -21,9 +21,10 @@ struct ocf_submit_volume_context {
 static void _ocf_volume_flush_end(struct ocf_io *io, int error)
 {
 	ocf_submit_end_t cmpl = io->priv1;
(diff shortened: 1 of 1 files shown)


arutk:
Hi @hammerg . We have identified this behavior as a defect in example application. OCF requires asynchronous queues implementation to make progress, not only in this specific case. Please see this change in example application:
https://github.com/Open-CAS/ocf/pull/575

hammerg:
> Hi @hammerg . We have identified this behavior as a defect in example application. OCF requires asynchronous queues implementation to make progress, not only in this specific case. Please see this change in example application: #575

Hi @arutk . Thanks for the update. I'll close this MR.

PR #651, opened 2022-01-24
[v21.3.4] Check superblock CRC before it is used 

arutk:


src/metadata/metadata_superblock.c
```diff
@@ -97,49 +98,138 @@ static void ocf_metadata_store_segment(ocf_pipeline_t pipeline,
 	ocf_pipeline_next(pipeline);
 }
 
-static void ocf_metadata_check_crc_sb_config(ocf_pipeline_t pipeline,
-		void *priv, ocf_pipeline_arg_t arg)
+
+#define ocf_log_invalid_meta(param) \
+	ocf_cache_log(cache, log_err, \
+			"Loading %s: invalid %s\n", \
+			ocf_metadata_segment_names[segment], \
```
mmichal10, line 105: How about using `metadata_segment_sb_config` instead of `segment` variable?

jfckm (approved):

src/metadata/metadata_superblock.c
arutk: done

mmichal10 (approved):

robertbaldyga (approved):

PR #655, opened 2022-02-02
Fix compilation error regarding stdarg.h file

hammerg:
Linux includes its own stdarg.h file since version 5.15. This change
allows ocf library to compile as part of a kernel module.

### DIFF
--- a/inc/ocf_logger.h
+++ b/inc/ocf_logger.h
@@ -11,8 +11,8 @@
  * @brief Logger API
  */
 
+#include <ocf_env.h>
 #include <ocf/ocf_types.h>
-#include <stdarg.h>
 
 /**
  * @brief Verbosity levels of context log


inc/ocf_logger.h
```diff
@@ -12,7 +12,17 @@
...
+#ifndef __KERNEL__
+	#include <stdarg.h>
+#else
+	#include <linux/version.h>
+	#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 15, 0)
+		#include <linux/stdarg.h>
+	#else
+		#include <stdarg.h>
+	#endif
+#endif
```
robertbaldyga, line 25: We want to keep OCF entirely platform agnostic. If there is any dependency (like this one) which can differ between different platforms, it should be rather handler through ocf_env.

robertbaldyga (changes requested):

inc/ocf_logger.h
hammerg: Did you mean something like this?

robertbaldyga: Yes. In posix env we can also drop `#ifndef __KERNEL__` as this is always userspace env. That kernel part should be implemented in kernel env defined in [Open CAS Linux ](https://github.com/Open-CAS/open-cas-linux) repository.

hammerg: Ok. I just replaced stdarg.h with ocf_env.h in ocf_logger and I'll submit a new patch for open-cas's env file.

robertbaldyga (approved):

robertbaldyga:
ok to test

arutk (approved):

PR #668, opened 2022-03-08
Handle a coredump when SPDK send a flush request

hammerg:
While trying to work with SPDK and OCF I had a core dump when a VM's disk is unmounted (due to FLUSH request). I created a test which reproduce this issue and suggested fixes.

Your comments are welcomed.

Thanks, Gal.

### DIFF
--- a/src/ocf_request.c
+++ b/src/ocf_request.c
@@ -105,7 +105,6 @@ struct ocf_request *ocf_req_new(ocf_queue_t queue, ocf_core_t core,
 		req->alloc_core_line_count = 1;
 	}
 
-
 	OCF_DEBUG_TRACE(cache);
 
 	ocf_queue_get(queue);
@@ -170,6 +169,26 @@ int ocf_req_alloc_map_discard(struct ocf_request *req)
 	req->core_line_last = ocf_bytes_2_lines(req->cache,
 			req->byte_position + req->byte_length - 1);
 	req->core_line_count = req->core_line_last - req->core_line_first + 1;
+
+	return ocf_req_alloc_map(req);
+}
+
+int ocf_req_alloc_map_flush(struct ocf_request *req)
+{
+	if (req->map)
+		return 0;
+
+	/*
+	 * Flush requests always apply to the whole disk. So in case the user
+	 * tries to submit an io request that failed to allocate mapping (e.g.
+	 * SPDK create a request with total disk size as request's length),
+	 * then allocate a small dummy mapping in order not to break future
+	 * actions on req->map.
+	 */
+	req->byte_length = 0;
+	req->core_line_first = 0;
+	req->core_line_last = 0;
+	req->core_line_count = 1;
 
 	return ocf_req_alloc_map(req);
 }
--- a/src/ocf_core.c
+++ b/src/ocf_core.c
@@ -361,6 +361,12 @@ static void ocf_core_volume_submit_flush(struct ocf_io *io)
 		return;
 	}
 
+	ret = ocf_req_alloc_map_flush(req);
+	if (ret) {
+		ocf_io_end(io, -OCF_ERR_NO_MEM);
+		return;
+	}
+
 	req->core = core;
 	req->complete = ocf_req_complete;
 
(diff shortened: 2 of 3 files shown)


robertbaldyga:
ok to test

src/ocf_request.c
```diff
@@ -97,14 +97,9 @@ struct ocf_request *ocf_req_new(ocf_queue_t queue, ocf_core_t core,
...
-		req->map = req->__map;
-		req->alock_status = (uint8_t*)&req->__map[core_line_count];
-		req->alloc_core_line_count = core_line_count;
-	} else {
-		req->alloc_core_line_count = 1;
-	}
-
+	req->map = req->__map;
+	req->alloc_core_line_count = map_allocated ? core_line_count : 1;
+	req->alock_status = (uint8_t*)&req->__map[req->alloc_core_line_count];
```
arutk, line 102: In standard I/O path whenever we fail to allocate map here, there is an attempt to allocate it from outside the memory pool (ocf_req_alloc_map called from ocf_core_volume_submit_io). After this change, ocf_req_alloc_map() will allways think that the map is allocated. This will result in out of bound access in case of large read/write I/O. It can easily be reproduced with a test similar to the one you added, just replace this one line
    core.submit_flush(0, core_device.size.bytes)
with
Rio().readwrite(ReadWrite.WRITE).bs(Size(core_device.size.bytes)).size(Size(core_device.size.bytes)).target(core).run()

arutk: I don't see any elegant way to fix this. I guess maybe flush should have its own map allocation logic, just like discard does (ocf_req_alloc_map_discard). This newly added ocf_req_alloc_map_flush would be called from ocf_core_volume_submit_flush and it would set req->map = req->__map in case req->map == NULL. This is good enough for flush (as it doesn't really need the map) and works with the case of large R/W I/O.

arutk (changes requested):

src/ocf_request.c
hammerg: I don't understand what you wrote regrading the failed allocation. In lines 91-94 I saw that if a "core_line_count" allocation from the pool fails then 1 core_line_count is allocated from the pool and not from outside of it.

arutk: Yes, in ocf_req_new() after failed allocation from the pool we attempt to alloc only 1 core line fro the pool. If this succeeds (the map_allocated = false case in line 100)  req->map remains NULL. Then in case of regular I/O (not discard) this results in ocf_req_alloc_map() calling env_zalloc() in an attempt to alloc multiple core lines separaterly from the request. 

On your branch the case that I describe above will have req->map != NULL. This will not trigger env_zalloc() allocation in ocf_req_alloc_map(). Further request processing will attempt to dereference req->map at indexes above 1, outside allocation bounds.

Are my instructions for the test clear to you? If so, you should be able to reproduce this easily in pyocf.

hammerg: Thanks for the extra information. I'll try to reproduce with your test case and see if I can propose a better solution.

Second222None:
I had the same problem. After applying this patch, an error occurs when copying a 3500 MB file into VM.
```
[2022-04-19 13:59:57.786267] Cache1.Aio2: Failed to write data to cache sector: 579584, bytes: 65536
[2022-04-19 13:59:57.808657] Cache1.Aio2: Failed to write data to cache sector: 594560, bytes: 65536
[2022-04-19 13:59:57.817270] Cache1.Aio2: Failed to write data to cache sector: 596352, bytes: 65536
[2022-04-19 13:59:57.834947] Cache1.Aio2: Failed to write data to cache sector: 611328, bytes: 65536
[2022-04-19 13:59:57.835303] Cache1.Aio2: Failed to write data to cache sector: 612992, bytes: 65536

```
 I don't know the code very well. Can anyone give some advice?

arutk (approved):

hammerg:
Can some please explain why the functional tests failed? I don't need any results in "Details". Thanks.

arutk:
This issue is properly fixed by not referencing map in flush engine: https://github.com/Open-CAS/ocf/pull/726

hammerg:
> This issue is properly fixed by not referencing map in flush engine: #726

Nice work. I'm closing this PR then.

PR #679, opened 2022-03-24
Don't flush and set shutdown bit while stopping standby cache

jfckm:

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/679)


jfckm:
This is not needed as standby has a separate stop pipeline, which doesn't set shutdown bit.

PR #687, opened 2022-03-30
Restore superblock copy after failed activation

arutk:


robertbaldyga (changes requested):
I proposed a better solution that prevents superblock modification during load / activate:
https://github.com/Open-CAS/ocf/pull/688

PR #699, opened 2022-04-11
Alleviation for static analyzers

mmichal10:


jfckm (dismissed):

src/cleaning/alru.c
```diff
@@ -517,7 +517,7 @@ static int ocf_alru_recovery_handle(ocf_parallelize_t parallelize,
 	int i;
 
 	portion = DIV_ROUND_UP((uint64_t)entries, shards_cnt);
-	begin = portion*shard_id;
+	begin = (uint64_t)portion*shard_id;
 	end = OCF_MIN(portion*(shard_id + 1), entries);
```
arutk, line 521: here such cast is actually required to avoid overflow in portion * (shard_id + 1)

src/cleaning/alru.c
```diff
@@ -517,7 +517,7 @@ static int ocf_alru_recovery_handle(ocf_parallelize_t parallelize,
 	int i;
 
 	portion = DIV_ROUND_UP((uint64_t)entries, shards_cnt);
-	begin = portion*shard_id;
+	begin = (uint64_t)portion*shard_id;
```
arutk, line 520: I would approach it differently. I'm not sure why begin and end are 64 bit in the first place. I would change then to 32bit unsigned.

arutk (dismissed):

arutk:
Sorry, I meant to request changes rather than approve

src/cleaning/acp.c
```diff
@@ -405,7 +405,7 @@ static void ocf_acp_recovery_chunk(struct ocf_acp_recovery_context *context,
 				.chunk[chunk->core_id][chunk->chunk_id];
 	}
 
-	for (bucket_id = 0; bucket_id < ACP_MAX_BUCKETS; bucket_id++) {
```
arutk, line 408: ignore this comment

arutk (changes requested):

robertbaldyga (approved):

PR #714, opened 2022-05-11
Copyright header check improvements

rafalste:
Make sure to look for all files that need to have copyright header included.

### DIFF
--- a/.github/verify_header.sh
+++ b/.github/verify_header.sh
@@ -5,20 +5,32 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
+# COPYRIGHT_REGEX is lowercase, because the whole line is
+# converted to lowercase before test against this regex.
+COPYRIGHT_REGEX="(copyright|\(c\))\s*([0-9]{4}(\s*-\s*([0-9]{4}))?)"
+LICENSE_REGEX="SPDX-License-Identifier: BSD-3-Clause$"
 YEAR=$(date +"%Y")
-REGEX="Copyright\(c\) [0-9]{4}-([0-9]{4}) |Copyright\(c\) ([0-9]{4}) "
 
-while read -r line; do
-    if [[ "$line" =~ $REGEX ]]; then
-        echo ${BASH_REMATCH[0]}
-        if [[ $YEAR == ${BASH_REMATCH[1]} || $YEAR == ${BASH_REMATCH[2]} ]]; then
-            echo $1 have appropriate license header
-            exit 0
-        fi
-        echo $1 have wrong license header year
-	exit 1
-    fi
+unset copyright_header license_header
+
+# Read lines until proper copyright and license headers are found.
+while read -r line && [[ ! "$copyright_header" || ! "$license_header" ]]; do
+	if [[ "${line,,}" =~ $COPYRIGHT_REGEX ]]; then
+		# If the fourth regex group (from year range) doesn't exist,
+		# use the second regex group instead (from a single year).
+		copyright_year=${BASH_REMATCH[4]:-${BASH_REMATCH[2]}}
+
+		if [[ $copyright_year == $YEAR ]]; then
+			copyright_header="correct_copyright_header_found"
+		fi
+	elif [[ "$line" =~ $LICENSE_REGEX ]]; then
+		license_header="correct_license_header_found"
+	fi
 done < "$1"
 
-echo $1 does not contain appropriate license header
+# Proper copyright and license info were found - all good.
+[[ "$copyright_header" && "$license_header" ]] && exit 0
+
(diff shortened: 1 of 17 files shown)


.github/workflows/pullrequest.yml
```diff
@@ -5,7 +5,8 @@ on:
       - master
 
 env:
-  EXTENSIONS: "c h cpp py go sh"
+  EXTENSIONS: "c h cpp py go sh mk spec service"
+  FILES: "*Makefile"
```
arutk, line 9: Maybe rename to "EXTENSIONLESS_FILES"?

arutk: Does this syntax with *asterisk work when Makefile is not in parent directory?

.github/workflows/pullrequest.yml
```diff
@@ -20,15 +21,16 @@ jobs:
...
-            if [[  "$FILE" =~ $REGEX ]]
-            then
-                EXTENSION=${BASH_REMATCH[1]}
-                EXTENSIONS_LIST=($EXTENSIONS)
-                if [[ " ${EXTENSIONS_LIST[*]} " =~ " ${EXTENSION} " ]]
-                then
-                    .github/verify_header.sh $FILE
-                fi
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            for file_in_list in $FILES; do
```
arutk, line 25: I think you can declare $FILES as an array rather than a string:
```
FILES=(*Makefile)
```
... and then explicitly check whether element is on the list instead of the inner loop:
```
if [[ " ${FILES[*]} " =~ " $file " ]]; then
```

.github/workflows/pullrequest.yml
rafalste: Yes, asterisk works. I've put it there because filenames we get from github are full paths, so asterisk is needed to cover all Makefiles in all directories.

And about EXTENSIONLESS_FILES... it is not entirely true. Well, it is in this case, but if you take a look at the same PR in OCL (https://github.com/Open-CAS/open-cas-linux/pull/1203) you can see that there are more files included in this variable, among which there are 'configure.d/*' - this directory contains files with .conf extension. We usually don't want config files to contain copyright header, but those files are an exception.
So I would say that in FILES we put all files that doesn't have an extension or they have but it's different from the one that we want to check every time... or any other reason. ;) So instead of FILES_WITHOUT_OR_WITH_DIFFERENT_EXTENSION I decided to name this just FILES. :D

.github/workflows/pullrequest.yml
rafalste: Yes I though about it also, but the reason I decided to make it like I did is how regular expression actually work here. Or more precisely - what is considered to be a wildcard here.
To match any Makefile we would need to specify it as `Makefile$` instead of `*Makefile`. While it is a small change here, it will be much more complicated in the same PR in OCL (https://github.com/Open-CAS/open-cas-linux/pull/1203/files) where we have more files included in this variable. For example `configure` should be put as `^configure$` to match only this file and not any file that has a 'configure' string in his full path name.
This could be done of course, but it will be much more complicated for future modifications while adding more files into it. I think this way is more user friendly. And this additional loop is not much of a performance overhead also, because we are testing there only files we put into FILES variable against files we modified in this PR (and not entire repo).

bkarpins:
valgrind this please

bkarpins:
valgrind this please

rafalste:
retest this please

rafalste:
retest this please

.github/verify_header.sh
```diff
@@ -5,20 +5,30 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
+COPYRIGHT_REGEX="Copyright\([Cc]\) ([0-9]{4}-([0-9]{4})|([0-9]{4}))[, ].*Intel Corporation$"
```
arutk, line 8: the check should work even if the copyright belongs to other company

.github/verify_header.sh
```diff
@@ -5,20 +5,30 @@
...
+
+    if [[ "$line" =~ $COPYRIGHT_REGEX ]]; then
+        if [[ ${BASH_REMATCH[2]} == $YEAR || ${BASH_REMATCH[3]} == $YEAR ]]; then
+            copyright_header=1
+            continue
         fi
-        echo $1 have wrong license header year
-	exit 1
+        copyright_header=0
+        echo >&2 "error: wrong year in copyright info: '${BASH_REMATCH[0]}'"
```
arutk, line 23: File may contain multiple "copyright" lines in random order, it's ok if any of them is up to date.

arutk (changes requested):

arutk (approved):

robertbaldyga (approved):

arutk (approved):

PR #716, opened 2022-05-16
Extend alock API with entries_count method

jfckm:
Right now alock assumes that number of locks taken will equal number of
core lines. This is not the case in pio, where only parts of metadata
are under locks. If pio request overlaps locked and not-locked metadata
section it will have different core lines number and awaited locks
number. To remedy this discrepancy additional method which gets count of
locks that will be taken/waited on is added to alock API.

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/716)

### DIFF
--- a/src/concurrency/ocf_cache_line_concurrency.c
+++ b/src/concurrency/ocf_cache_line_concurrency.c
@@ -1,5 +1,5 @@
 /*
- * Copyright(c) 2012-2021 Intel Corporation
+ * Copyright(c) 2012-2022 Intel Corporation
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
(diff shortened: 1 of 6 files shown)


src/concurrency/ocf_cache_line_concurrency.c
```diff
@@ -141,9 +141,17 @@ static int ocf_cl_lock_line_slow(struct ocf_alock *alock,
...
+		struct ocf_request *req)
+{
+	return req->core_line_count;
+}
+
 static struct ocf_alock_lock_cbs ocf_cline_conc_cbs = {
 		.lock_entries_fast = ocf_cl_lock_line_fast,
-		.lock_entries_slow = ocf_cl_lock_line_slow
+		.lock_entries_slow = ocf_cl_lock_line_slow,
+		.get_entries_count = ocf_cl_get_entries_count
```
arutk, line 153: ocf_cl_get_entries_count -> ocf_cl_lock_get_entries_count?

rafalste (approved):

arutk (dismissed):

src/concurrency/ocf_pio_concurrency.c
```diff
@@ -136,9 +136,26 @@ static int ocf_pio_lock_slow(struct ocf_alock *alock,
...
 
+static uint32_t ocf_pio_lock_get_entries_count(struct ocf_alock *alock,
+		struct ocf_request *req)
+{
+	uint32_t i, count = 0;
+	ocf_cache_line_t entry;
+
+	for (i = 0; i < req->core_line_count; i++) {
+		entry = ocf_pio_lock_get_entry(alock, req, i);
+		if (entry == OUT_OF_RANGE)
```
mmichal10, line 147: I guess the majority of requests will be modifying entries that are in range. What do you think about adding `unlikely()`?

mmichal10 (dismissed):

mmichal10 (approved):

arutk (approved):

PR #719, opened 2022-05-16
Extend alock API with entries_count method

jfckm:
Right now alock assumes that number of locks taken will equal number of
core lines. This is not the case in pio, where only parts of metadata
are under locks. If pio request overlaps locked and not-locked metadata
section it will have different core lines number and awaited locks
number. To remedy this discrepancy additional method which gets count of
locks that will be taken/waited on is added to alock API.

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/ocf/719)


arutk (approved):

mmichal10:
Fixes #728

PR #724, opened 2022-05-18
pyocf: flush engine logic test

arutk:


tests/functional/tests/engine/test_flush.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def trace_flush(vol, io, io_type):
+        nonlocal flushes
+
+        if io_type == TraceDevice.IoType.Flush:
+            if not vol.uuid in flushes:
+                flushes[vol.uuid] = []
+            flushes[vol.uuid].append((io.contents._addr, io.contents._bytes))
+
+        return True
```
rafalste, line 29: Is this intended to always return True? `TraceDevice._trace()` seem to rely on the return value from this `trace_flush()`.

jfckm: Return from trace function decides whether or not to let the IO through, in this case I think Adam didn't mean to inject any errors so `return True` seems to be apropriate

rafalste (dismissed):

jfckm (approved):

rafalste (approved):

jfckm (approved):

jfckm:
@CAS-Linux-Jenkins retest this please

PR #732, opened 2022-06-14
Validate metadata segments in cleaning section - Pipeline conditional step

pdebski21:


src/metadata/metadata.c
```diff
@@ -1118,20 +1118,42 @@ static void ocf_metadata_load_all_finish(ocf_pipeline_t pipeline,
...
+static bool ocf_metadata_check_disabled_segment(ocf_pipeline_t pipeline,
+		void* priv, ocf_pipeline_arg_t arg)
+{
+	struct ocf_metadata_context *context = priv;
+	int segment_id = ocf_pipeline_arg_get_int(arg);
+	if (context->ctrl->segment[segment_id] != NULL)
+		return true;
+	return false;
+
+}
```
robertbaldyga, line 1142: I'm not sure if we really want to check for NULL segment (which is kind of side effect rather than actual API).
We could name this function `ocf_check_if_cleaner_disabled` (which is way more self-explaining and context-providing) and check the actual cache configuration.

I know where the temptation to make it more generic comes from, as we generate some amount of boilerplate, seemingly for nothing. But in any modern programming language the predicate would be a lambda function instead, and it would be as specific to the given context as possible (which in general is a good practice). So I'd rather follow the design pattern, even at cost of the boilerplate, rather than providing future-compatible, hackish filter function.

src/utils/utils_pipeline.c
```diff
@@ -39,6 +39,12 @@ static int _ocf_pipeline_run_step(struct ocf_request *req)
 			pipeline->next_step++;
 			step->hndl(pipeline, pipeline->priv, &step->arg);
 			return 0;
+		case ocf_pipeline_step_conditional:
+			arg = &step->args[pipeline->next_arg++];
```
robertbaldyga, line 43: Step conditional should take only one arg.

src/utils/utils_pipeline.c
```diff
@@ -39,6 +39,12 @@ static int _ocf_pipeline_run_step(struct ocf_request *req)
...
+			if (step->cond(pipeline, pipeline->priv, arg)) {
+				step->hndl(pipeline, pipeline->priv, arg);	
+				return 0;
+			}
```
robertbaldyga, line 47: We need to increment `pipeline->next_step++`, otherwise in the next iterations we will try to execute the same step over and over again.

src/utils/utils_pipeline.h
```diff
@@ -112,6 +117,14 @@ struct ocf_pipeline_step {
 		.type = ocf_pipeline_step_terminator, \
 	}
 
+#define OCF_PL_STEP_CONDITIONAL(_hndl, _args, _pred) \
+	{ \
+		.type = ocf_pipeline_step_conditional, \
+		.hndl = _hndl, \
+		.args = _args, \
```
robertbaldyga, line 124: Again, it should take a single arg, not an array.

robertbaldyga (changes requested):

src/metadata/metadata.c
pdebski21: I see, checking whether loading metadata cleaning section is needed through checking if cleaner disabled option is set
can be better way to go. Good point.

src/metadata/metadata.c
```diff
@@ -1118,20 +1118,37 @@ static void ocf_metadata_load_all_finish(ocf_pipeline_t pipeline,
...
 	OCF_PL_ARG_INT(metadata_segment_list_info),
 	OCF_PL_ARG_INT(metadata_segment_hash),
 	OCF_PL_ARG_TERMINATOR(),
 };
 
+
+/*
+ * Predicate function checking whether disable cleaner option is set
+ */
+static bool ocf_check_if_cleaner_disabled(ocf_pipeline_t pipeline,
```
robertbaldyga, line 1132: So now it's rather `ocf_check_if_cleaner_enabled`.

src/metadata/metadata.c
```diff
@@ -1118,20 +1118,37 @@ static void ocf_metadata_load_all_finish(ocf_pipeline_t pipeline,
...
+		void* priv, ocf_pipeline_arg_t arg)
+{
+	struct ocf_metadata_context *context = priv;
+	return !context->cfg->disable_cleaner;
```
robertbaldyga, line 1136: Missing empty line between declarations and code.

src/utils/utils_pipeline.h
```diff
@@ -112,6 +117,14 @@ struct ocf_pipeline_step {
 		.type = ocf_pipeline_step_terminator, \
 	}
 
+#define OCF_PL_STEP_CONDITIONAL(_hndl, _arg, _pred) \
+	{ \
+		.type = ocf_pipeline_step_conditional, \
+		.hndl = _hndl, \
+		.arg = _arg, \
```
robertbaldyga, line 124: It's a bit more complicated. See `OCF_PL_STEP_ARG_INT`/`OCF_PL_STEP_ARG_PTR` for reference.

robertbaldyga (changes requested):

robertbaldyga (dismissed):

bkarpins:
valgrind this please

bkarpins:
valgrind this please

src/utils/utils_pipeline.c
```diff
@@ -39,6 +39,11 @@ static int _ocf_pipeline_run_step(struct ocf_request *req)
 			pipeline->next_step++;
 			step->hndl(pipeline, pipeline->priv, &step->arg);
 			return 0;
+		case ocf_pipeline_step_conditional:
+			pipeline->next_step++;
+			if (step->pred(pipeline, pipeline->priv, &step->arg))
+				step->hndl(pipeline, pipeline->priv, &step->arg);	
+			return 0;
```
arutk, line 46: I think you should add "continue" if the predicate returns false.

arutk (dismissed):

src/metadata/metadata.c
```diff
@@ -1118,20 +1118,37 @@ static void ocf_metadata_load_all_finish(ocf_pipeline_t pipeline,
 
 struct ocf_pipeline_arg ocf_metadata_load_all_args[] = {
```
arutk, line 1119: also should do the same ocf_metadata_flush_all_args

src/utils/utils_pipeline.c
```diff
@@ -39,6 +39,11 @@ static int _ocf_pipeline_run_step(struct ocf_request *req)
 			pipeline->next_step++;
 			step->hndl(pipeline, pipeline->priv, &step->arg);
 			return 0;
+		case ocf_pipeline_step_conditional:
+			pipeline->next_step++;
+			if (step->pred(pipeline, pipeline->priv, &step->arg))
+				step->hndl(pipeline, pipeline->priv, &step->arg);	
+			continue;
```
arutk, line 46: you should "continue" only if predicate returns false. otherwise return 0.

robertbaldyga (approved):

jfckm (approved):

PR #740, opened 2022-06-20
utils: submit a whole disk flush request

hammerg:
A flush request should include both starting address and number of
blocks to flush. Since the ocf_submit_volume_flush function doesn't
include parameters we assume the whole disk should be flushed.

### DIFF
--- a/src/utils/utils_io.c
+++ b/src/utils/utils_io.c
@@ -30,8 +30,10 @@ void ocf_submit_volume_flush(ocf_volume_t volume,
 		ocf_submit_end_t cmpl, void *priv)
 {
 	struct ocf_io *io;
+	uint64_t length;
 
-	io = ocf_volume_new_io(volume, NULL, 0, 0, OCF_WRITE, 0, 0);
+	length = ocf_volume_get_length(volume);
+	io = ocf_volume_new_io(volume, NULL, 0, length, OCF_WRITE, 0, 0);
 	if (!io)
 		OCF_CMPL_RET(priv, -OCF_ERR_NO_MEM);
 


jfckm:
@CAS-Linux-Jenkins ok to test

arutk:
Hi @hammerg. What is your motivation for requiring address & size in flush request? 

ocf_submit_volume_flush is obviously meant to flush the entire disk. It does so by sending I/O without data and with FLUSH flag set. So far we assumed FLUSH flag to behave similarly to REQ_PREFLUSH in Linux kernel. In this context there is no need for address/size, as REQ_PREFLUSH flushes entire device volatile cache. Address/size in bio with REQ_PREFLUSH determines data payload target LBA, and there is no data in this case.

hammerg:
The motivation comes from two reasons:

1. SPDK crash when you're trying to use OCF with a raid0 bdev as a cache volume. The raid bdev is aware to the fact that the flush request have no payload but fails due to the fact that the blocks number is zero.
2. Since the SCSI spec in the SYNCHRONIZE CACHE (10) command allow the NUMBER OF BLOCKS field to be set to zero and I couldn't decide whatever the fix should be in OCF or SPDK ;-)

Your call.

Thanks, Gal.

jfckm:
Hi Gal,

Thanks for thorough explanation. I've checked out SCSI spec section you mentioned - indeed, what we do is within specification and I believe fix in SPDK is necessary. However, it is also true that our understanding of flush was lacking.

We plan on releasing this Friday, so we won't take on additional work at this time. We don't plan another release until September and if I properly understand SPDKs release cycle any improvements in our 22.9 won't make it to SPDK until 23.1 - that may be a motivating factor for fix in SPDK too ;)

Anyways, thanks for bringing up this issue and providing a patch, but I think that in this particular instance we won't merge it  - we'd like to keep environment-specific changes to minimum, especially since it's SPDK raid bdev that's out of spec here.

Cheers,
jfckm

hammerg:
Hi Jan,

Thanks for review and the honest reply and I agree with your decision. I'll try fix this issue in SPDK.

    Gal.

PR #746, opened 2022-08-17
Stack memory reduction for OCF stats

pdebski21:
Fixes Open-CAS/open-cas-linux#1212
Signed-off-by: Piotr Debski <piotr.debski@intel.com>

### DIFF
--- a/src/ocf_stats_builder.c
+++ b/src/ocf_stats_builder.c
@@ -1,10 +1,11 @@
 /*
- * Copyright(c) 2012-2021 Intel Corporation
+ * Copyright(c) 2012-2022 Intel Corporation
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
 #include "ocf/ocf.h"
 #include "ocf_priv.h"
+#include "ocf_env.h"
 #include "metadata/metadata.h"
 #include "engine/cache_engine.h"
 #include "utils/utils_user_part.h"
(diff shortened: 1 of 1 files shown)


kmajzero (dismissed):
checked on Ubuntu 20.04 with a custom debug kernel. No warnings or errors found. LGTM!

src/ocf_stats_builder.c
```diff
@@ -446,13 +455,17 @@ int ocf_stats_collect_cache(ocf_cache_t cache,
...
 
 	if (blocks)
-		_fill_blocks(blocks, &s);
+		_fill_blocks(blocks, s);
 
 	if (errors)
-		_fill_errors(errors, &s);
+		_fill_errors(errors, s);
 
 	return 0;
```
rafalste, line 466: and here

src/ocf_stats_builder.c
```diff
@@ -346,32 +349,37 @@ int ocf_stats_collect_core(ocf_core_t core,
...
-		_fill_errors(errors, &s);
-
+		_fill_errors(errors, s);
+	
+	env_vfree(s)
 	return 0;
+
+collect_failed:
+	env_vfree(s);
+	return result;
```
rafalste, line 382: Actually, since 'result' must equal 0 at all times for this function to complete successfully, I would remove the first free and return statements and leave only those after 'collect_failed' mark.

rafalste: And you may also rename collect_failed to someting like mem_free maybe? This will be less confusing during normal non-error flow.

rafalste (approved):

src/ocf_stats_builder.c
pdebski21: Good point It was in my mind for a moment. Will be changed :)

rafalste (approved):

kmajzero (approved):

jfckm (approved):

PR #757, opened 2022-10-01
alru: add parameter `max_dirty_ratio'

live4thee:
With a high dirty ratio and occupancy, OCF might unable to map cache lines for new requests, thus pass-through the I/O to core devices.  IOPS will drop afterwards.  Existing `alru' policy gives user the chance to control the stale buffer time, activity threshold etc.  They can affect the dirty ratio of the cache device, but in an empirical manner, more or less.  Introducing `max_dirty_ratio' can make it explicit.

At first glance, it might be better to implement a dedicated cleaner policy directly targeting dirty ratio goal, so that the `alru' parameters remains orthogonal.  But one the other hand, we still need to flush dirty cache lines periodically, instead of just keeping a watermark of dirty ratio. It indicates that existing `alru' parameters are still required if we develop a new policy, and it seems reasonable to make it a parameter.

To sum up, this patch does the following:
- added a 'max_dirty_ratio' parameter with default value 100;
- with default value 100, `alru' cleaner is identical to what is was;
- with value N less than 100, the cleaner (when waken up) will active brought dirty ratio to N, regardless of staleness time.

I  have tested this patch with a modified Open-CAS and it works. If any comments, please feel free.

### DIFF
--- a/example/simple/src/main.c
+++ b/example/simple/src/main.c
@@ -94,7 +94,7 @@ int initialize_cache(ocf_ctx_t ctx, ocf_cache_t *cache)
 	ocf_mngt_cache_config_set_default(&cache_cfg);
 	cache_cfg.metadata_volatile = true;
 
-	/* Cache deivce (volume) configuration */
+	/* Cache device (volume) configuration */
 	type = ocf_ctx_get_volume_type(ctx, VOL_TYPE);
 	ret = ocf_uuid_set_str(&uuid, "cache");
 	if (ret)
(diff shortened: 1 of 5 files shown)


robertbaldyga (approved):
@live4thee Thanks for the contribution! The change looks good to me.
Please add your company (or your personal) copyright line in headers of files you changed.

kmajzero (approved):

kmajzero (approved):
Please update the copyright header at the beginning of alru.h with a correct date.

PR #758, opened 2022-10-05
Add diskseen prefetch framework.

Second222None:
Add DISKSEEN ([DiskSeen: exploiting disk layout and access history to enhance I/O prefetch]) prefetch framework.
The data structure used for recording the read histroy is a block table. It is inspired by the multi-level page table. In this patch, we use a two-level block table. It consumes 2MiB of memory and can record history for about 512MiB file access working set.  The prefetch routine is invoked when there are more than 4 sequential reads. The prefetch length is (8*io->bytes).

Signed-off-by: Song Wei <second222none@163.com>


kmajzero (changes requested):

robertbaldyga:
[To Github Support] I authorize removal of this Pull Request.

PR #763, opened 2022-11-04
Avoid evicting target partition

DocentSzachista:
In case when there are ioclasses with the same eviction priority and target ioclass is first on the ioclass list, then the target ioclass is being evicted. To avoid this situation make sure to skip it when choosing ioclass to evict.

Signed-off-by: Damian Raczkowski <damian.raczkowski@intel.com>

### DIFF
--- a/src/ocf_space.c
+++ b/src/ocf_space.c
@@ -59,6 +59,9 @@ static inline uint32_t ocf_evict_user_partitions(ocf_cache_t cache,
 		/*
 		 * Check stop and continue conditions
 		 */
+		if (req->part_id == part_id)
+            continue;
+
 		if (max_priority > user_part->config->priority) {
 			/*
 			 * iterate partition have higher priority,


mmichal10:
Please test the following scenario:
- create a small cache (i.e. 100M) with a significantly bigger core (500M) in WT and seq-cutoff=never
- populate the cache with a simple dd if=/dev/zero of=/dev/cas1-1 bs=1M count=120M
- call dd for the second time, but to a different rage of LBAs dd if=/dev/zero of=/dev/cas1-1 bs=1M count=120 seek=200
- check cache statistic for pass-through IO. If there is any, this patch introduces workload malformations thus needs to be redesigned

DocentSzachista:
> Please test the following scenario:
> 
> * create a small cache (i.e. 100M) with a significantly bigger core (500M) in WT and seq-cutoff=never
> * populate the cache with a simple dd if=/dev/zero of=/dev/cas1-1 bs=1M count=120M
> * call dd for the second time, but to a different rage of LBAs dd if=/dev/zero of=/dev/cas1-1 bs=1M count=120 seek=200
> * check cache statistic for pass-through IO. If there is any, this patch introduces workload malformations thus needs to be redesigned

Provided scenario does not generate pass-through IO.

PR #773, opened 2023-03-30
[v22.6.1] Fix alignment of private data in parallelize & pipeline

robertbaldyga:
there is an issue when someone call to parallelize/pipeline with some struct that is aligned (say to 64B)
but these APIs add their own data, right before
the user's private data.
so, the user's data is no longer aligned
which might cause segfault in some cases.

### DIFF
--- a/src/utils/utils_parallelize.c
+++ b/src/utils/utils_parallelize.c
@@ -1,5 +1,6 @@
 /*
  * Copyright(c) 2012-2021 Intel Corporation
+ * Copyright(c) 2023 Huawei Technologies
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
@@ -8,6 +9,8 @@
 #include "../engine/engine_common.h"
 #include "../ocf_request.h"
 #include "utils_parallelize.h"
+
+#define OCF_PARALLELIZE_ALIGNMENT 64
 
 struct ocf_parallelize {
 	ocf_cache_t cache;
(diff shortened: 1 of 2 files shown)


mmichal10 (dismissed):

arutk:
I think this patch doesn't to much to improve alignment of the private data.

What it changes is it assures that the private data starts at an offset (relative to master allocation) which is a multiplicity of 64 bytes. The actual master allocation alignment remains at whatever the default of env_vzalloc is (env_vzalloc is not aware of alignment requirements). 

Before this change priv pointer was guaranteed to be aligned to the smaller of the two:
1. vzalloc default alignment
2. depending on which structure we're considering:
    a. struct ocf_pipeline alignment
    b. min(ocf_paralellize alignment, pointer alignment) - as there is also some space saved for an array of  ocf_request pointers between the end of ocf_parallelize struct and beginning of the priv space.

This change assures that 2.a and 2.b are multiplicity of 64, so effectively the priv is now aligned to min(vzalloc default alignment, 64). So it somewhat improves the alignment of priv assuming (for the case of ocf_pipeline_t) that the master structure ocf_pipeline_t alignment requirements are lower than vzalloc default. 

If your intention is to align priv to 64 then one of the following woud work:
1. just add extra 64 bytes to both env_vzalloc allocations and explicitly align the pointer to priv upon assignment. 
2. keep your changes but also make sure that whatever comes from master allocation is 64-byte aligned. This is harder as you would need to either 
a. change allocation method so that allocation alignment can be specified
b. again inflate the master allocation by 64 bytes and manually increment the pointer so that it is rounded to 64 bytes - but this time the master pointer, no the priv.

I assumed that vzalloc has some fixed default alignment. I imagine it can somehow implicitly deduce alignment requirement from the size parameter, but depending on this would probably not be portable.

arutk (approved):

Issue #794, opened 2024-06-13
If SECURE_MEMORY_HANDLING is enabled, env_secure_free() fails to clear freed memory


wfrisch:
## Description
`ocf_env.h` has an optional SECURE_MEMORY_HANDLING mode that's supposed to clear freed memory chunks in env_secure_free().

This function is called by:
```
src/metadata/metadata_hash.c
683:    env_secure_free(context, sizeof(*context));

src/metadata/metadata_raw.c
81:             env_secure_free(raw->mem_pool, raw->mem_pool_limit);

src/metadata/metadata_raw_dynamic.c
129:            env_secure_free(ctrl->pages[i], PAGE_SIZE);
300:    env_secure_free(context->page, PAGE_SIZE);
```

## Expected Behavior
env_secure_free() should clear freed memory

## Actual Behavior
env_secure_free() does not clear freed memory

## Possible Fix
Fix the argument order of memset() in env_secure_free().

wfrisch:
Fixed by https://github.com/Open-CAS/ocf/pull/796

PR #795, opened 2024-06-13
env: posix: fix memset args in env_secure_free()

wfrisch:
The argument order was wrong. See also:
https://bugzilla.suse.com/show_bug.cgi?id=1226132

Fixes issue #794


wfrisch:
Duplicate of https://github.com/Open-CAS/ocf/pull/796

PR #798, opened 2024-07-05
Add missing ocf_cache_mode_t to ocf_req_cache_mode_t conversions

robertbaldyga:


rafalste (approved):
Is there a need for such distinction if it's just type casting?

robertbaldyga:
> Is there a need for such distinction if it's just type casting?

It's not just type casting. Value set of `ocf_req_cache_mode_t` is superset of `ocf_cache_mode_t`, which just so happens to have the same modes mapped to the same numerical values in both. This property is however an implementation detail (so that mapping is trivial), and it should not leak to the actual usages of both types.

mmichal10 (approved):

PR #801, opened 2024-07-25
Revert "cleaner: Remove complete_queue"

robertbaldyga:
This functionality is used by cleaning policies via cmpl_queue to reschedule the completion, so that we avoid unlocking mutex in the cleaner completion from interrupt context of IO completion.

This reverts commit 1e5eda68a7e1abb7e9c6bdfa6c5d2a4f940681dc.

### DIFF
--- a/src/utils/utils_cleaner.c
+++ b/src/utils/utils_cleaner.c
@@ -104,6 +104,7 @@ static struct ocf_request *_ocf_cleaner_alloc_master_req(
 	/* In master, save completion context and function */
 	req->priv = attribs->cmpl_context;
 	req->master_io_req = attribs->cmpl_fn;
+	req->complete_queue = attribs->cmpl_queue;
 
 	/* The count of all requests */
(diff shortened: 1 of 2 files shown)


src/utils/utils_cleaner.c
```diff
@@ -203,12 +215,18 @@ static void _ocf_cleaner_complete_req(struct ocf_request *req)
...
+	} else {
+		/* Only master contains completion function and priv */
+		cmpl = master->master_io_req;
+		cmpl(master->priv, master->error);
 
-	/* For additional get on master allocation */
-	ocf_req_put(master);
+		/* For additional get on master allocation */
+		ocf_req_put(master);
+	}
```
mmichal10, line 229: I wonder how much of an optimization would be
```
if (req != master && master->complete_queue)
    push_req_front()
else
    cmpl()
```
to avoid redundant locking

mmichal10 (approved):

rafalste (approved):

Issue #636, opened 2022-01-03
ocf_io_set_data() does not accept offset != 0


arutk:
## Description
OCF returns -OCF_ERR_INVAL if offset != 0 is provided to ocf_io_set_data() - both for core and cache volume I/O.

## Expected Behavior
offset != 0 is accepted

mmichal10:
#816 fixed the issue for core volumes only

PR #816, opened 2024-09-09
Support set_data() with offset > 0

mmichal10:


mmichal10:
Fixes #636

rafalste (approved):

robertbaldyga (approved):

mmichal10:
Requires #824

tests/functional/tests/basic/test_offset.py
```diff
@@ -0,0 +1,63 @@
...
+    io = core_volume.new_io(queue, addr, length, IoDir.WRITE, 0, 0)
+    io.set_data(data, offset)
+    Sync(io).submit()
+
+    # normal read
+    s = core_volume.read_sync_4k(queue, 4096, len(B2))
+    assert s == B2
+
+    s = core_device.read_sync_4k(queue, 4096, len(B2))
+    assert s == B2
```
Deixx, line 48: is this repeated read needed?

tests/functional/tests/basic/test_offset.py
```diff
@@ -0,0 +1,63 @@
...
+    io1.set_data(data1, offset1)
+    Sync(io1).submit()
+
+    s0 = data1.buffer[:offset1]
+    assert s0 == bytes([Data.DATA_POISON] * offset1)
+    s1 = data1.buffer[offset1:(offset1 + len(B2))]
+    assert s1 == B2
+
+    core_volume.close()
+    cache.stop()
```
rafalste, line 63: Isn't test-prepare or some other mechanism handle this closing and stopping?

tests/functional/tests/basic/test_offset.py
```diff
@@ -0,0 +1,63 @@
...
+    data.seek(DataSeek.BEGIN, 0)
+    data.write(B1, len(B1))
+    data.seek(DataSeek.BEGIN, 4096)
+    data.write(B2, len(B2))
+
+    addr = 4096
+    length = 4096
+    offset = 4096
+    io = core_volume.new_io(queue, addr, length, IoDir.WRITE, 0, 0)
+    io.set_data(data, offset)
```
rafalste, line 40: Shouldn't the addr or offset be 0 if we want to perform a normal read?

tests/functional/tests/basic/test_offset.py
mmichal10: First, we read from the core and then from the backed device so it actually makes sense :)

tests/functional/tests/basic/test_offset.py
mmichal10: Not quite. The test prepares an 8k buffer but instead of writing its whole content, it writes only its second half by setting offset to 4k.

tests/functional/tests/basic/test_offset.py
mmichal10: done

rafalste (approved):

robertbaldyga (approved):

PR #826, opened 2024-09-13
Errors in engines

mmichal10:
Requires #816


jfckm (approved):

rafalste (approved):

robertbaldyga (dismissed):

src/engine/engine_wt.c
```diff
@@ -117,16 +112,20 @@ static void _ocf_write_wt_req_complete(struct ocf_request *req)
...
 }
 
 static void _ocf_write_wt_cache_complete(struct ocf_request *req, int error)
 {
 	if (error) {
+		/* Cache error code is not propagated further to the user here
+		 * because data could be successfully written to the core device
+		 * despite the cache IO error.
+		 * Error flag is set though to indicate that the error occurred
+		 * and to invalidate the request in completion. */
```
Deixx, line 127: If we don't want to propagate the cache error code to the user, is line 129 required? I understand this ensures `req->error` is always != 0 and the check in line 132 is obsolete.

src/ocf_request.h
```diff
@@ -192,6 +195,12 @@ struct ocf_request {
 	int error;
 	/*!< This filed indicates an error for OCF request */
 
+	int cache_error;
+	/*!< Indicator of forward IO cache device error */
+
+	int core_error;
+	/*!< Indicator of forward IO core device error */
+
```
Deixx, line 203: I'm trying to find where these are used

robertbaldyga: In the next patchset ;)

src/engine/engine_wt.c
robertbaldyga: Fixed.

Deixx (approved):

rafalste (approved):

PR #827, opened 2024-09-13
engine_rd refactor

mmichal10:
Requires #826

### DIFF
--- a/src/engine/engine_rd.c
+++ b/src/engine/engine_rd.c
@@ -28,32 +28,23 @@ static void _ocf_read_generic_hit_complete(struct ocf_request *req, int error)
 	struct ocf_alock *c = ocf_cache_line_concurrency(
 			req->cache);
 
+	OCF_DEBUG_RQ(req, "HIT completion");
+
 	if (error) {
 		req->error |= error;
 		ocf_core_stats_cache_error_update(req->core, OCF_READ);
 		inc_fallback_pt_error_counter(req->cache);
 	}
 
-	/* Handle callback-caller race to let only one of the two complete the
-	 * request. Also, complete original request only if this is the last
-	 * sub-request to complete
-	 */
-	if (env_atomic_dec_return(&req->req_remaining) == 0) {
-		OCF_DEBUG_RQ(req, "HIT completion");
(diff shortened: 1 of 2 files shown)


rafalste (approved):

src/engine/engine_rd.c
```diff
@@ -28,78 +28,65 @@ static void _ocf_read_generic_hit_complete(struct ocf_request *req, int error)
...
-		if (req->error) {
-			/*
-			 * --- Do not submit this request to write-back-thread.
-			 * Stop it here ---
-			 */
-			req->complete(req, req->error);
+	if (req->error) {
+		/*
+		 * --- Do not submit this request to write-back-thread ---
+		 */
```
jfckm, line 66: Submit to what? :D

src/engine/engine_rd.c
```diff
@@ -106,21 +106,25 @@ static inline void _ocf_read_generic_submit_miss(struct ocf_request *req)
...
+err_alloc:
 	/* Submit read request to core device. */
 	ocf_submit_volume_req(&req->core->volume, req,
 			_ocf_read_generic_miss_complete);
-
-	return;
-
-err_alloc:
-	_ocf_read_generic_miss_complete(req, -OCF_ERR_NO_MEM);
 }
```
jfckm, line 128: Something's wrong in here. Previously when alloc/mlock failed we completed with error and submited request if neither failed. Now we always submit request and never complete with error?

jfckm (changes requested):

src/engine/engine_rd.c
mmichal10: If the buffer allocation fails we can't backfill but we it's not a reason to fail the user's IO

src/engine/engine_rd.c
mmichal10: Once upon the time... 😆

jfckm (approved):

Deixx (commented):
requires rebase, but otherwise lgtm

Deixx (approved):

jfckm (approved):

PR #831, opened 2024-09-18
Io forward pt2

mmichal10:
Requires #829

### DIFF
--- a/src/engine/engine_pt.c
+++ b/src/engine/engine_pt.c
@@ -79,7 +79,7 @@ int ocf_read_pt_do(struct ocf_request *req)
 	ocf_engine_update_block_stats(req);
 
 	ocf_core_stats_pt_block_update(req->core, req->part_id, req->rw,
-			req->byte_length);
+			req->bytes);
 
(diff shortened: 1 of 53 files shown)


src/engine/engine_pt.c
```diff
@@ -79,7 +79,10 @@ int ocf_read_pt_do(struct ocf_request *req)
 	ocf_engine_update_block_stats(req);
 
 	ocf_core_stats_pt_block_update(req->core, req->part_id, req->rw,
-			req->byte_length);
+			req->bytes);
+
+	ocf_core_stats_pt_block_update(req->core, req->part_id, req->rw,
+			req->bytes);
```
Deixx, line 85: duplicated lines above

Deixx, line 85: should it be called twice?

mmichal10: It shouldn't, nice catch :)

Deixx (approved):

robertbaldyga (approved):

PR #834, opened 2024-09-20
Backfill optimization

mmichal10:


Deixx (dismissed):

rafalste (approved):
Seems OK. My only consideration is, if it wouldn't be better to use the existing implementation of ocf_engine_forward_cache_io_req() and just add the conditional `if (req->map[i].status == LOOKUP_HIT)` to it (as a flag maybe), instead of copying over the whole function content.

mmichal10:
@rafalste I think the implementation is too specific for backfill to apply it in a generic function like `ocf_engine_forward_cache_io_req()`. The hits couldn't be skipped for read requests so it would generate another special case to handle. Also, the optimizatized code doesn't actually skip all hits; it skips a hit only if it's not preceded by an adjacent miss (it's to avoid sending too many small requests to cache if the cachelines are hits and misses in turn)

src/engine/engine_bf.c
```diff
@@ -68,11 +68,21 @@ static void _ocf_backfill_complete(struct ocf_request *req, int error)
 	}
 }
 
+#define __entries_not_adjacent(__req, __i, __j) \
+	__req->map[__i].coll_idx + 1 != __req->map[__j].coll_idx
+
+#define __is_the_last_chunk(__req, __i) (__i == (__req->core_line_count - 1))
+
+#define __skip_on_the_last_entry(__skip, __req, __i) \
+	__skip * (int)__is_the_last_chunk(__req, __i)
```
robertbaldyga, line 77: It's no longer 90's, `unlikely()` should do.

src/engine/engine_bf.c
```diff
@@ -96,49 +106,40 @@ static int _ocf_backfill_do(struct ocf_request *req)
 		return 0;
 	}
 
+	seek = req->addr % cache_line_size;
+	last_chunk_size = (req->addr + req->bytes) % cache_line_size;
+	skip = (cache_line_size - last_chunk_size) % cache_line_size;
```
robertbaldyga, line 111: This math is very confusing. The `last_chunk_size` is not really last chunk size, because it can be 0 if last chunk is aligned to cache line size, and then to account for that, the `skip` is "corrected" with this surprising modulo.

src/engine/engine_bf.c
```diff
@@ -96,49 +106,40 @@ static int _ocf_backfill_do(struct ocf_request *req)
...
-			 */
-			addr_next = 0;
-			total_bytes += bytes;
+			/* This is the 1st cache line in the interval, and it's
+			   a hit. Don't write it to the cache */
+			total_bytes += cache_line_size;
+			total_bytes -= seek;
+			/* Seek should be taken into account for the first chunk
+			   only */
+			seek = 0;
```
robertbaldyga, line 122: I'd prefer `unlikely()` on `if (i == 0)`. Now we are zeroing this `seek` here and there many times for completely no reason.

src/engine/engine_bf.c
```diff
@@ -96,49 +106,40 @@ static int _ocf_backfill_do(struct ocf_request *req)
...
 
-		if (i == (req->core_line_count - 1)) {
-			skip = (ocf_line_size(cache) -
-					((req->addr + req->bytes) %
-					 ocf_line_size(cache))) % ocf_line_size(cache);
-			bytes -= skip;
-		}
+		/* Seek should be taken into account for the first chunk only */
+		addr += seek;
+		bytes -= seek;
```
robertbaldyga, line 139: And here we completely unnecessarily add/subtract 0 on every loop iteration.

robertbaldyga (changes requested):

PR #847, opened 2024-10-04
Refactors&fixes

mmichal10:


Deixx (approved):

example/simple/src/main.c
```diff
@@ -360,49 +355,61 @@ int main(int argc, char *argv[])
...
+		printf("Error: Unable to remove core\n");
+		goto core_err;
+	}
 
 	/* Stop cache */
 	ocf_mngt_cache_stop(cache1, simple_complete, &context);
 	sem_wait(&context.sem);
-	if (ret)
-		error("Unable to stop cache\n");
+	if (ret) {
```
rafalste, line 396: 'ret' is not set for this check

src/mngt/ocf_mngt_cache.c
rafalste, line 1: Duplicated commit message. Maybe use "Log errors on cache start" or squash with the previous one.

example/simple/src/main.c
mmichal10: The address of `ret` is assigned to `context.error` in line 362 so the value would be updated by `simple_complete()`

rafalste (approved):

PR #856, opened 2024-10-16
Debug request tracing

mmichal10:


src/engine/cache_engine.c
```diff
@@ -118,8 +133,29 @@ static const struct ocf_io_if *cache_mode_io_if_map[ocf_req_cache_mode_max] = {
...
 
+void ocf_debug_request_trace(struct ocf_request *req,
+		ocf_req_cache_mode_t engine, uint8_t info)
+{
+	uint8_t tmp = (uint8_t)engine;
+
+	ENV_BUG_ON(info > 0xf);
+	ENV_BUG_ON(engine > 0xf);
+
+	ENV_WARN_ON(req->engine_trace & 0xff00000000000000);
```
robertbaldyga, line 150: `ENV_WARN_ONCE()` should be enough for all of those checks..

src/engine/cache_engine.c
```diff
@@ -118,8 +133,29 @@ static const struct ocf_io_if *cache_mode_io_if_map[ocf_req_cache_mode_max] = {
...
+	[ocf_req_cache_mode_rd] = &IO_IFS[OCF_IO_RD_IF],
 };
 
+void ocf_debug_request_trace(struct ocf_request *req,
+		ocf_req_cache_mode_t engine, uint8_t info)
+{
+	uint8_t tmp = (uint8_t)engine;
+
+	ENV_BUG_ON(info > 0xf);
+	ENV_BUG_ON(engine > 0xf);
```
robertbaldyga, line 148: We are working with bitmasks here, so maybe, to keep it consistent, `& ~0xf` ?

src/engine/engine_common.c
```diff
@@ -599,6 +599,8 @@ static int _ocf_engine_refresh(struct ocf_request *req)
 {
 	int result;
 
+	ocf_debug_request_trace(req, ocf_req_refresh, 0);
+
```
robertbaldyga, line 603: Having the tracking on the entry point to an engine loses information about the original source of the request. For example, there are three places where req can be redirected from engine_rd to engine_pt, but tracking only the entry point we don't know exactly which path has been taken.

The better solution would be to have this tracking on the redirect point. So for all the request the first redirect point would be `ocf_engine_hndl_req()` or `ocf_engine_hndl_fast_req()` (and then we automatically know if it's fast path or not, even if it's not an engine_fast!). Then we need to find all the other redirect points within the engines, and call the tracing function from there.

robertbaldyga (changes requested):

PR #860, opened 2025-01-17
Fix error messages for metadata found during attach

Deixx:


inc/ocf_mngt.h
```diff
@@ -438,6 +438,14 @@ struct ocf_mngt_cache_attach_config {
...
 	 */
 	bool disable_cleaner;
+
+	/**
+	 * @brief Indicates that attach-cache command was called
+	 *
+	 * Used for proper error handling to differentiate between start-cache
+	 * and attach-cache operations, and display proper error message.
+	 */
+	bool is_cmd_attach;
```
robertbaldyga, line 448: This is an implementation detail and should not be exposed to the adapter.

robertbaldyga (changes requested):

inc/ocf_mngt.h
Deixx: Sure, I've changed it to utilise cache_state

Deixx: I've moved the differentiation to OCL, but this small change is still required

Deixx:
Metadata will be zeroed on detach

PR #867, opened 2025-03-17
Detach vs acp

mmichal10:
Fixes https://github.com/Open-CAS/open-cas-linux/issues/1621


mmichal10:
The bug doesn't reproduce for cache stop because before cleaning policy deinit, all cores had already been removed from metadata and from system.  
For detach, the core related metadata had been deinitialized, but the core was still present in the system, so the cleaning_deinit() was trying to deinitialize it again.

robertbaldyga (approved):

rafalste (approved):

PR #868, opened 2025-03-24
Fix occupancy accounting

mmichal10:


mmichal10:
Regarding checkpatch's warning:
image
The dereference is too long. I tried to fit it in one line but then it exceeded 80 characters.

Deixx (approved):

jfckm (approved):

src/utils/utils_cache_line.c
```diff
@@ -20,20 +20,26 @@ static void __set_cache_line_invalid(struct ocf_cache *cache, uint8_t start_bit,
...
-		if (line_became_invalid) {
-			env_atomic_dec(&core->runtime_meta->cached_clines);
-			env_atomic_dec(&core->runtime_meta->
-					part_counters[part_id].cached_clines);
-		}
-		ocf_lru_rm_cline(cache, line);
-		ocf_metadata_remove_cache_line(cache, line);
+	if (line_remains_valid) {
+		ENV_BUG_ON(line_became_invalid);
+		return;
```
robertbaldyga, line 25: This `ENV_BUG_ON()` is a new addition here, and mixing it into the logic decreases readability.
I'd rather keep it simple and let the compiler do its thing:
```c
ENV_BUG_ON(line_became_invalid && line_became_invalid);

if (line_remains_valid)
	return;
```

PR #881, opened 2025-04-16
Repartition data after loading IO class configuration

Deixx:


src/ocf_lru.c
```diff
@@ -588,6 +592,69 @@ static void ocf_lru_clean_end(void *private_data, int error)
...
+			cline = ocf_metadata_get_lru(cache, cline)->prev;
+		}
+
+		if (cline != end_marker) {
+			core_id = ocf_metadata_get_core_id(cache, cline);
+			core = ocf_cache_get_core(cache, core_id);
+			if (metadata_test_dirty(cache, cline))
+				ocf_cleaning_purge_cache_block(cache, cline);
+
+			ocf_lru_repart_locked(cache, cline, part, dst_part);
```
mmichal10, line 628: I don't think `ocf_lru_repart_locked()` is a good fit for repartition during a managment operation because every reparted cache line will be marked as hot and will trigger unnecessary rebalancing on the default ioclass.

src/mngt/ocf_mngt_io_class.c
```diff
@@ -328,3 +329,30 @@ int ocf_mngt_cache_io_classes_configure(ocf_cache_t cache,
...
+	int i;
+
+	/* Temporarily disable IO classes (except for default) for
+	 * repartitioning.
+	 */
+	for (i = 1; i < OCF_USER_IO_CLASS_MAX; i++)
+		_ocf_mngt_set_partition_size(cache, i, 0, 0);
+
+	for (i = 1; i < OCF_USER_IO_CLASS_MAX; i++) {
+		ocf_lru_repart_all(cache,
```
mmichal10: Why do we need to repart all ioclasses?

src/ocf_lru.c
```diff
@@ -781,8 +871,7 @@ void ocf_lru_hot_cline(ocf_cache_t cache, ocf_cache_line_t cline)
 	if (hot)
 		return;
 
-	part_id = ocf_metadata_get_partition_id(cache, cline);
-	part = &cache->user_parts[part_id].part;
+	part = &cache->user_parts[node->partition_id].part;
```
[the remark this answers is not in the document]
Deixx: Indeed :)

src/mngt/ocf_mngt_io_class.c
Deixx: The 3rd commit in this PR adds an optional parameter that decides whether only non-active or all partitions should undergo the process.

PR #882, opened 2025-04-24
fix: Check if cache attached when switching cleaning policy

jfckm:


src/mngt/ocf_mngt_flush.c
```diff
@@ -1067,6 +1067,12 @@ void ocf_mngt_cache_cleaning_set_policy(ocf_cache_t cache,
 	if (ocf_cache_is_standby(cache))
 		OCF_CMPL_RET(priv, -OCF_ERR_CACHE_STANDBY);
 
+	if (!ocf_cache_is_device_attached(cache)) {
+		ocf_cache_log(cache, log_err, "Cannot change cleaning policy - "
+				"cache device is detached\n");
```
Deixx, line 1072: doesn't extended_err_msg in casadm take care of the error message?
```
	{
		OCF_ERR_CACHE_DETACHED,
		"The operation is not permited while the cache is detached"
	},
```

robertbaldyga: It does, but OCL is not the only user of OCF.

robertbaldyga (approved):

mmichal10 (approved):

src/mngt/ocf_mngt_flush.c
Deixx: OK, but it is inconsistent with other instances of returning this (and other) errors. Also in OCL this will display both the messages, and they are not complementing, just repeating in different words.

PR #887, opened 2025-06-05
[RFC] Changes for the new SPDK OCF module.

rafalste:


rafalste:
Checkpatch CI check reports "Possible repeated word" warning which is a false positive and can be ignored. Repeated word is just a documentation syntax.

robertbaldyga (approved):

PR #893, opened 2025-09-22
Sort flush data only if there is anything to flush

rafalste:
If there is nothing to sort, the env_sort() is called with NULL pointer instead of the first element of array to sort.
This could be fixed in specific implementation of env_sort() to return immediately if number of elements is 0 but all other routines check the element count before calling ocf_cleaner_sort_flush_data() except of ocf_cleaner_sort_flush_containers(), so putting this check there seems more appropriate and makes flush data sorting less prone to errors.

### DIFF
--- a/src/utils/utils_cleaner.c
+++ b/src/utils/utils_cleaner.c
@@ -837,8 +837,9 @@ void ocf_cleaner_sort_flush_containers(struct flush_container *fctbl,
 	int i;
 
 	for (i = 0; i < num; i++) {
-		ocf_cleaner_sort_flush_data(fctbl[i].flush_data,
-				fctbl[i].count);
+		if (fctbl[i].count > 1)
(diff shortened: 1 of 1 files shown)


src/utils/utils_cleaner.c
```diff
@@ -837,8 +837,9 @@ void ocf_cleaner_sort_flush_containers(struct flush_container *fctbl,
 	int i;
 
 	for (i = 0; i < num; i++) {
-		ocf_cleaner_sort_flush_data(fctbl[i].flush_data,
-				fctbl[i].count);
+		if (fctbl[i].count > 0)
```
mmichal10, line 840: If there is only one element, sorting is not necessary, so technically, the condition could be `fctbl[i].count > 1` 😆

mmichal10 (approved):

PR #894, opened 2025-09-22
Improve core flush

rafalste:


src/mngt/ocf_mngt_flush.c
```diff
@@ -265,8 +265,10 @@ static int _ocf_mngt_get_flush_containers(ocf_cache_t cache,
 	}
 
 	if (!dirty_total) {
-		_fcnum = 0;
+		*fcnum = 0;
 		goto free_fc;
+	} else {
```
robertbaldyga, line 270: `else` not needed here, because the `if` statement does `goto`.

robertbaldyga (changes requested):

mmichal10 (approved):

PR #960, opened 2026-06-19
Get cleaner from cache

rafalste:
Needed by SPDK OCF module to set/get CPU mask on cleaner thread.

### DIFF
--- a/src/ocf_cache.c
+++ b/src/ocf_cache.c
@@ -288,6 +288,22 @@ ocf_ctx_t ocf_cache_get_ctx(ocf_cache_t cache)
 	return cache->owner;
 }
 
+ocf_cleaner_t ocf_cache_get_cleaner(ocf_cache_t cache)
+{
+	OCF_CHECK_NULL(cache);
+
+	if (ocf_cache_is_standby(cache))
+		return NULL;
+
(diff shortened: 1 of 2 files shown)


src/ocf_cache.c
```diff
@@ -288,6 +288,19 @@ ocf_ctx_t ocf_cache_get_ctx(ocf_cache_t cache)
...
 
+ocf_cleaner_t ocf_cache_get_cleaner(ocf_cache_t cache)
+{
+	OCF_CHECK_NULL(cache);
+
+	if (!ocf_cache_is_device_attached(cache))
+		return NULL;
+
+	if (cache->conf_meta->cleaner_disabled)
+		return NULL;
```
robertbaldyga, line 302: Also this (and doc update):
```c
if (ocf_cache_is_standby(cache))
	return NULL;
```
BTW this function this is the first place where we introduce `ocf_cleaner_t` nullability, but I think API-wise it is ok.

robertbaldyga (changes requested):

src/ocf_cache.c
rafalste: Any other doc for this change then the one in header?

robertbaldyga (approved):

PR #963, opened 2026-07-04
Eliminate C23 extensions

rafalste:
SPDK doesn't allow using C23 extensions and throws an error in the upstream CI during compilation:
"error: '_Static_assert' with no message is a C23 extension [-Werror,-Wc23-extensions]"

Add a message as a second argument for _Static_assert() to avoid using C23 extensions.


rafalste:
Checkpatch complains about a line too long in a commit description. This line is an error message and wrapping it would make it less readable in my opinion. Please merge as is if possible.

